PR #8, opened 2016-04-29
Remove useless dependences

yyjdelete:
1. Fix build for `DynamicLinqWebDocs`.
2. Dynamic load `System.Data.Entity`or`EntityFramework` as predefined types if it is found.
3. `SqlFunctions`, `SqlSpatialFunctions` and `DbFunctions`(EF6) is also supported.
4. Allow to overwrite keywords.
5. Allow to use FullName of type(***May not works***).
6. `System.Data.Entity` is always here, so overwrite short name of it with `EntityFramework` if `EntityFramework` is found. If `System.Data.Entity` is needed, FullName of type can be used(***FullName may not works as 5***).
7. Remove usless dependences(`System.Data.Entity`,`JetBrains.Annotations` and many others for dnxcore5 and dotnet54).

### DIFF
--- a/src/System.Linq.Dynamic.Core/ExpressionParser.cs
+++ b/src/System.Linq.Dynamic.Core/ExpressionParser.cs
@@ -282,27 +282,29 @@ namespace System.Linq.Dynamic.Core
             try
             {
                 Type efType;
-                //EF5, System.Data.Objects.DataClasses.EdmFunctionAttribute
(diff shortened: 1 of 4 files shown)


StefH:
Thanks for this pull-request, however, I cannot figure out how to create a unit-test to test this change ? Do you have any ideas?

PR #37, opened 2016-07-20
Support strings as Enum Parameter Objects

jogibear9988:


jogibear9988:
I can only compile net 4.5 on my system, and this works. hopefully you can help to fix the other?

StefH:
1]
There is a build error:
`C:\projects\system-linq-dynamic-core\src\System.Linq.Dynamic.Core\ExpressionParser.cs(678,44): error CS7036: There is no argument given that corresponds to the required formal parameter 'ignoreCase' of 'Enum.Parse(Type, string, bool)'`

2]
Can you please add/update unit-tests ?

jogibear9988:
Done

StefH:
Enum Parse with 'true' is case sensitive; is that what we want? I think other things are case insensitive?

jogibear9988:
sorry, i wanted to use false.


Von meinem iPhone gesendet

StefH:
Merged and new NuGet created : https://www.nuget.org/packages/System.Linq.Dynamic.Core/1.0.5.2

PR #38, opened 2016-07-23
Support more comparisons with strings

jogibear9988:


jogibear9988:
Sorry, I could not create unit tests, because I could not compile it on my system:

I get this error:

The .NET Core SDK is not installed or cannot be found under the path C:\Program Files\dotnet\bin. These components are required to build and run this project. Download the version of .NET Core SDK specified in global.json or update the SDK version in global.json to the version that is installed.

jogibear9988:
Maybe you could help me, so I can fix the compilation errors and create tests.

StefH:
You need latest vs2015 update 3.
Also need .net core rc2 sdk. And maybe also .net core tooling for vs2015.


See this linl https://www.microsoft.com/net

jogibear9988:
I needed to uninstall a old Version, then it worked.

StefH:
some comments
* TypeConverter is present for silverlight, so I'll enable this
* The dependencies for older frameworks on System.ComponentModel.TypeConverter is not needed, it's already present.

I'll fix above items and add code to new branch, you can review/test this and if you agree, I merge back to main.

jogibear9988:
I've seen you changed some unit tests using left and right as value. I think we should nearly modify all tests to compare both sides in ExpressionParser

StefH:
Indeed, surely for new functionality we should add left and right unit-tests.

PR #55, opened 2016-12-05
Fix Nullable Enums from String

jogibear9988:


StefH:
Can you update/create a unit-test for this specific issue?

StefH:
Thank you !

Issue #66, opened 2017-04-01
Is there Way to enter a Complex query


jogibear9988:
For example, I've a IEnumerable Object and would like to give this string to dynamic linq:

     "GroupBy(x=>x.Id).OrderBy(x=>x.Key).Select(x=>x.Key).Where(x=>x!=null)"

for Example smth like this
  
      aaa.Complex("GroupBy(x=>x.Id).OrderBy(x=>x.Key).Select(x=>x.Key).Where(x=>x!=null)")

and it returns a object, so the result could be everything

StefH:
Maybe this one can help you?
https://github.com/StefH/System.Linq.Dynamic.Core/blob/10fc9a2ec8677f638210804c0ca636de021e6c9a/test/System.Linq.Dynamic.Core.Tests/ComplexTests.cs

jogibear9988:
No. I want everything in one string.

jogibear9988:
You see here: https://dynamiclinq.codeplex.com/SourceControl/latest#DynamicLinq/DynamicLinq/Program.cs
look at the second query

StefH:
Ah ok. I am not sure on this, if my fork does even support this notation. Did you try this?

jogibear9988:
I'll try... If not, I'll look to patch that this will also work!

StefH:
Thanks a lot.

jogibear9988:
Does not work ... :-( Now we need to look what to do...

jogibear9988:
One problem: In my commit I removed the 

     Type type = _resultType;

Problem here is, we should only do this on the last "new" in the string. How can we decide if this is the case?

jogibear9988:
I've now added it back, but it is a Problem if have have two select new statements, where the first one should be an anonymous type!

StefH:
I dont have access to my pc this week. Please investigate some more, I will try to look and merge at sunday. (which branch did you take?)

jogibear9988:
fixed

StefH:
@jogibear9988 

Can you create a unit-test which tests all this code?

```c#
Expression ParseLambdaOperator()
        {
            Expression expr = ParseOrOperator();
            if (_textParser.CurrentToken.Id == TokenId.Lambda && _it.Type == expr.Type)
            {
                _textParser.NextToken();
                if (_textParser.CurrentToken.Id == TokenId.Identifier ||
                    _textParser.CurrentToken.Id == TokenId.OpenParen)
[...]

PR #68, opened 2017-04-02
Work on #66 -> Should work now. Tests will follow on VS2017 support!

jogibear9988:


jogibear9988:
Why is appveyor red?

StefH:
- build ok
- tests ok
- coverage ok
- uploading to coveralls fails because token is not set it seems. I will fix in the weekend

PR #85, opened 2017-05-31
Fix Guid? == null comparison

jogibear9988:
Can you create a new nuget?

### DIFF
--- a/src/System.Linq.Dynamic.Core/ExpressionParser.cs
+++ b/src/System.Linq.Dynamic.Core/ExpressionParser.cs
@@ -83,12 +83,8 @@ namespace System.Linq.Dynamic.Core
             void F(Guid? x, Guid? y);
             void F(Guid x, string y);
             void F(Guid? x, string y);
-            void F(Guid x, object y);
-            void F(Guid? x, object y);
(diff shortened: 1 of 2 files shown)


StefH:
Unit tests fail. Can you take a look at your code?

jogibear9988:
I thought, maybe we should work on 

int FindBestMethod(IEnumerable<MethodBase> methods, Expression[] args, out MethodBase method)
        
the different created arrays seem not verry perfomant to me

jogibear9988:
fixed...

jogibear9988:
would be nice if it could be merged, and you could create a new nuget

jogibear9988:
can you create a new nuget?

StefH:
I did merge, and added some extra tests.

However it some tests fails? Not related to guid I think, maybe I missed something...
See:
https://github.com/StefH/System.Linq.Dynamic.Core/blob/master/test/System.Linq.Dynamic.Core.Tests/ExpressionTests.cs#L583

jogibear9988:
I fixed your test, you need to use 

      new object[] {null}

but there was also a error then in the lib, because the expression is not the same as the cached null expression!

PR #103, opened 2017-09-09
support group by with 2 parameters, add tolist

jogibear9988:


jogibear9988:
and what about this?

src/System.Linq.Dynamic.Core/ExpressionParser.cs
```diff
@@ -181,6 +182,7 @@ interface IEnumerableSignatures
             void LastOrDefault();
             void Single();
             void SingleOrDefault();
+            void ToList();
```
StefH: Is there a test to cover this method?

StefH: And a function like `ToArray`, can that also be defined?

test/System.Linq.Dynamic.Core.Tests/DynamicExpressionParserTests.cs
```diff
@@ -54,6 +54,48 @@ public void ParseLambda_EmptyParameterList()
         }
 
         [Fact]
+        public void ParseLambdaComplex_1()
```
StefH: Is it possible to move this test to the file QueryableTests.GroupBy.cs or does that not fit?

StefH:
I did add some question to this code-review.

test/System.Linq.Dynamic.Core.Tests/DynamicExpressionParserTests.cs
```diff
@@ -54,6 +54,48 @@ public void ParseLambda_EmptyParameterList()
...
+            var expected = qry.GroupBy(x => new { x.Profile.Age }, x => x).OrderBy(gg => gg.Key.Age).Select(j => new { j.Key.Age, TotalIncome = j.Sum(k => k.Income) }).Select(c => new ComplexParseLambda1Result { Age = c.Age, TotalIncome = c.TotalIncome }).Cast<dynamic>().ToArray();
+
+            // Assert
+            Check.That(result).IsNotNull();
+            Check.That(result).HasSize(expected.Length);
+            Check.That(result.ToArray()[0]).Equals(expected[0]);
+        }
+
+        [Fact]
+        public void ParseLambdaComplex_2()
```
jogibear9988: This tests ToList()

src/System.Linq.Dynamic.Core/ExpressionParser.cs
jogibear9988: Yes in ParseLambdaComplex_2

jogibear9988: Think so, but I've not tested/used

test/System.Linq.Dynamic.Core.Tests/DynamicExpressionParserTests.cs
jogibear9988: The other tests in this class (for example ParseLambda_1) also test Lambda with GroupBy

Issue #136, opened 2018-01-08
Expressions on dynamic objects


NickDarvey:
Hi,

I'm wondering what the story is for `System.Linq.Dynamic.Core` and types that aren't known till runtime. Is this supported right now, and if it's not, will it be one day?

Right now, I have an `IEnumerable<dynamic>` I can that I'd like to query. Is there any way I can compose a query on top of that? Is there any other way you know of for handling this case?
(The collection contains `ExpandoObject`s from a JSON payload deserialized by Json.NET. I see you have a test for `JObjects` but we don't want to be bound to that type as we may have multiple serializers.)

There is a test in there for verifying a `Select` on a collection of dynamic objects, but it's disabled right now.

# Repro
[Run this test](https://github.com/StefH/System.Linq.Dynamic.Core/blob/02fd83b24b0e8be92659aabec9f6ae45382f2ae2/test/System.Linq.Dynamic.Core.Tests/ExpressionTests.cs#L1217). 
```c#
//[Fact]
public void ExpressionTests_Select_ExpandoObjects()
{
    //Arrange
    dynamic a = new ExpandoObject();
    a.Name = "a";
    a.BlogId = 100;
    dynamic b = new ExpandoObject();
    b.Name = "b";
    b.BlogId = 100;
    var list = new List<dynamic> { a, b };
    IQueryable qry = list.AsQueryable();

    var result = qry.Select("it").Select("BlogId");

    //Assert
    Assert.Equal(new[] { 100, 200 }, result.ToDynamicArray<int>());
}
```

# Expected
Passes.

# Actual
It's commented out. It fails when running it:
>Message: System.Linq.Dynamic.Core.Exceptions.ParseException : No property or field 'BlogId' exists in type 'Object'

StefH:
Also related to https://github.com/StefH/System.Linq.Dynamic.Core/issues/132

StefH:
A possible solution could be that a new class is defined in this library which keeps the information about the properties.
And the properties can be defined with the help from https://github.com/StefH/System.Linq.Dynamic.Core/blob/10fc9a2ec8677f638210804c0ca636de021e6c9a/src/System.Linq.Dynamic.Core/DynamicProperty.cs

So prototype code would be:
``` cs
var property1 = new DynamicProperty("BlogId", typeof(int));
var property2 = new DynamicProperty("Name", typeof(string));
var a = new DynamicObject(property1, property2);
a.Name = "a";
[...]

PR #137, opened 2018-01-09
Feature: Add support for querying a IQueryable<dynamic>

NickDarvey:
Updates the `ParseMemberAccess` method so that if it can't get the property or field, it tries getting the member by a dynamic operation or by an indexer operation.

After a lot of searching I ended up on [this bit of code](https://github.com/dkackman/SqlLinq/blob/210b594e37f14061424397368ed750ce547c21e7/SqlLinq/SyntaxTree/ExpressionFactory.cs#L114) which seems to do the trick. The `ExpressionTests_Select_ExpandoObjects` test passes now.

Some notes:
- It will throw an exception only during evaluation if it can't find the specified member (and it is an `object` or has an indexer), currently a `InvalidOperationException` but perhaps introducing a new exception type is warranted?
- I'd still like to get `ExpressionTests_Select_DynamicObjects` but I'm not sure how I'll go.
- @StefH: Are you able to pull this and run all the tests on your machine? I was only able to get the solution building with a lot of hacking of projects and stripping out all platforms except net461.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,6 +1,6 @@
 os: Visual Studio 2017
 
-version: 1.0.7.{build}
+version: 1.0.8.{build}
 
 configuration:
 - Debug
--- a/src-console/ConsoleAppEF1.1/Program.cs
+++ b/src-console/ConsoleAppEF1.1/Program.cs
@@ -1,5 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Dynamic;
+using System.Linq;
 using System.Linq.Dynamic.Core;
 using Newtonsoft.Json;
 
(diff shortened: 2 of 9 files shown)


NickDarvey:
Oh sweet! CI! Okay, will take a look at which platform is dying...

NickDarvey:
Oh, also tagging #136

NickDarvey:
Ah, okay, it's failing to build `System.Linq.Dynamic.Core.Tests.csproj` because `System.Linq.Expressions.Expression.Dynamic Method (CallSiteBinder, Type, IEnumerable<Expression>)` is not available in `netcoreapp1.1`

What do you think about changing the test project to target `netcoreapp2.0` (which has the method is available) and make this feature conditional and only available for `netstandard2.0` platforms?

StefH:
I'll take your PR and take a look.

NickDarvey:
Alright, `ExpressionTests_Select_DynamicObjects` might need to come in another PR... Can this get merged in without it?

StefH:
I will merge this code (at this point) to main and create a new NuGet. The discussion on other possible solutions will be done into the issue [https://github.com/StefH/System.Linq.Dynamic.Core/issues/136].

PR #158, opened 2018-04-18
Fix157

jogibear9988:


jogibear9988:
will create a better fix, wich does not use the tostring of the parameter

StefH:
No problems. But i need some time to check these config changes and probably also change the version from nuget because its a breaking API change

jogibear9988:
When you look here: https://codeblog.jonskeet.uk/2018/04/13/backward-compatibility-and-overloading/ you'll see, it's really hard to do a change wich will not break in every situation.

But normaly my changes should not be breaking changes, cause I always also created also the Extension Method without the parameter!

PR #166, opened 2018-05-15
Feature: Added support for implicit type conversions

arjenvrh:
Added support for implicit type conversions (i.e. a class defines an implicit casting operator to the type of a ConstantExpression).

```csharp
public class ImplicitCastingClass
{
    public Guid Id { get; set; }
    public static implicit operator Guid(ImplicitCastingClass id)
    {
        return id.Id;
    }
}
```


chait-01:
I am using dynamic lambda expression, and trying to filter records based on a derived type.

var expr1 = DynamicExpressionParser.ParseLambda<Settings, bool>(new ParsingConfig(), true, "c => (c.Applicability As WinApplicability).SKUs.Contains(@0)", result);

Here Appilicability is a property within Settings type, and WinApplicability is a derived type of Applicability. SKUs is a property within WinApplicability.

The above line of code fails with the below error. Any idea how to deal with this issue?

Issue #19, opened 2016-05-20
NotEqual filter not working with DateTime


Muchiachio:
```C#
public class Program
{
    public static void Main(string[] args)
    {
        IQueryable<Model> models = new[] { new Model() }.AsQueryable();

        Console.WriteLine(models.Where("Date != null").ToArray().Length);
    }
}

public class Model
{
    public DateTime Date { get; set; }
}
```

Throws `The binary operator NotEqual is not defined for the types 'System.DateTime' and 'System.String'`. This works in `System.Linq.Dynamic` and `System.Linq`.

StefH:
Issue is solved, some unit-tests are also added.

Can you please verify this library ? Or do you want a NuGet to test ?

Muchiachio:
I confirmed, it's working.

PR #200, opened 2018-08-23
Fix for parsing Guid and string in the same condition

OlegNadymov:
Plus fix for parsing expression with custom types with implicit convertions


OlegNadymov:
Hi @StefH !
Could you show me more information about failing unit test and code issues.
All unit test were passed on my local machine.

test/System.Linq.Dynamic.Core.Tests/DynamicExpressionParserTests.cs
```diff
@@ -452,5 +454,39 @@ public void ParseLambda_With_InnerStringLiteral()
             object result = del.DynamicInvoke(String.Empty);
             Check.That(result).IsEqualTo(originalTrueValue);
         }
+      
+        [Fact]
+        public void ParseLambda_With_If_Guid_Null()
```
StefH: Please use `// Arrange` , `// Act` and `// Assert` flow (take a look at other tests).

test/System.Linq.Dynamic.Core.Tests/DynamicExpressionParserTests.cs
```diff
@@ -452,5 +454,39 @@ public void ParseLambda_With_InnerStringLiteral()
...
+        {
+            var objContext = new User();
```
StefH: user

test/System.Linq.Dynamic.Core.Tests/DynamicExpressionParserTests.cs
```diff
@@ -452,5 +454,39 @@ public void ParseLambda_With_InnerStringLiteral()
...
+            var guidEmpty = Guid.Empty;
```
StefH: For clarity, I mostly / always use real definitions for simple types. So use `Guid` here.
Only for complex types, I use var.

test/System.Linq.Dynamic.Core.Tests/DynamicExpressionParserTests.cs
```diff
@@ -452,5 +454,39 @@ public void ParseLambda_With_InnerStringLiteral()
...
+            var someId = Guid.NewGuid();
+            var expressionText = $"iif(@0.{nameof(objContext.Id)} == null, @0.{nameof(objContext.Id)} == Guid.Parse(\"{someId}\"), {nameof(objContext.Id)}={nameof(objContext.Id)})";
```
StefH: string

test/System.Linq.Dynamic.Core.Tests/DynamicExpressionParserTests.cs
StefH: Shouldn't this be `Iif` ?

test/System.Linq.Dynamic.Core.Tests/DynamicExpressionParserTests.cs
StefH: For readability I would just type the expressionText string as simple as possible, so just:
``` c#
$"iif(@0.Id == null, @0.Id == Guid.Parse(\"{someId}\"), Id=Id)";
```

And actually I wonder if this is ok?
Wouldn't the last statement set a value?
Like:
``` c#
$"iif(@0.Id == null, @0.Id == Guid.Parse(\"{someId}\"), Id)";
```

test/System.Linq.Dynamic.Core.Tests/DynamicExpressionParserTests.cs
StefH: Better text like: `Guid_Equals_Null` ?

test/System.Linq.Dynamic.Core.Tests/DynamicExpressionParserTests.cs
```diff
@@ -452,5 +454,39 @@ public void ParseLambda_With_InnerStringLiteral()
...
+            var boolLambda = lambda as Expression<Func<User, bool>>;
+            Check.That(boolLambda).IsNotEqualTo(null);
+
+            var del = lambda.Compile();
+            object result = del.DynamicInvoke(objContext);
+            Check.That(result).IsEqualTo(true);
+        }
+
+        [Fact]
+        public void ParseLambda_With_If_Null_Guid()
```
StefH: Same comments as defined on other method also apply here.

src/System.Linq.Dynamic.Core/Parser/SupportedOperands/IEqualitySignatures.cs
```diff
@@ -13,9 +13,13 @@ internal interface IEqualitySignatures : IRelationalSignatures
...
-        void F(Guid x, string y);
-        void F(Guid? x, string y);
-        void F(string x, Guid y);
-        void F(string x, Guid? y);
+
+        // Disabled 2 lines below according to unit test ParseLambda_With_If_Guid_Null
+        //void F(Guid x, string y);
+        //void F(Guid? x, string y);
+
+        // Disabled 2 lines below according to unit test ParseLambda_With_If_Null_Guid
```
StefH: I would think that removing this code would actually break unit-tests? Or do you say that this logic was wrong from the beginning?

OlegNadymov:
@StefH,
I can't open the link https://www.codefactor.io/repository/github/stefh/system.linq.dynamic.core/pull/200.
I see the error:
"Access is denied. You do not have permission to access /repository/github/stefh/system.linq.dynamic.core/pull/200."

src/System.Linq.Dynamic.Core/Parser/SupportedOperands/IEqualitySignatures.cs
OlegNadymov: In my opinion logic was wrong from the start.
It's like the issue #19
I've added unit test ``ParseLambda_With_Guid_Equals_String`` for demonstrate that ``Guid`` plus ``String`` works without these lines (at least in condition part).

test/System.Linq.Dynamic.Core.Tests/DynamicExpressionParserTests.cs
OlegNadymov: It was done!

test/System.Linq.Dynamic.Core.Tests/DynamicExpressionParserTests.cs
OlegNadymov: It was done!

test/System.Linq.Dynamic.Core.Tests/DynamicExpressionParserTests.cs
OlegNadymov: It was done!

test/System.Linq.Dynamic.Core.Tests/DynamicExpressionParserTests.cs
OlegNadymov: It was done!

test/System.Linq.Dynamic.Core.Tests/DynamicExpressionParserTests.cs
OlegNadymov: It was done!

OlegNadymov: It was done!

test/System.Linq.Dynamic.Core.Tests/DynamicExpressionParserTests.cs
OlegNadymov: It was done!

OlegNadymov:
@StefH all changes according to your review were done!

StefH:
[CodeFactor](https://www.codefactor.io/repository/github/stefh/system.linq.dynamic.core) found multiple issues last seen at 6213fd1790d5ed57e703734195f4e9bf50ec0f55:

#### An opening curly bracket must not be followed by a blank line.
[test\System.Linq.Dynamic.Core.Tests\DynamicExpressionParserTests.cs:13
](https://github.com/StefH/System.Linq.Dynamic.Core/blob/6213fd1790d5ed57e703734195f4e9bf50ec0f55/test/System.Linq.Dynamic.Core.Tests/DynamicExpressionParserTests.cs#L13)
#### The code must not contain multiple blank lines in a row.
[test\System.Linq.Dynamic.Core.Tests\DynamicExpressionParserTests.cs:15
](https://github.com/StefH/System.Linq.Dynamic.Core/blob/6213fd1790d5ed57e703734195f4e9bf50ec0f55/test/System.Linq.Dynamic.Core.Tests/DynamicExpressionParserTests.cs#L15)

StefH:
Small issues:
- Please use `// Arrange` , `// Act` and `// Assert` flow (take a look at other tests). What I mean is that you add comment lines to make easy to read. See for example this: https://github.com/StefH/System.Linq.Dynamic.Core/blob/master/test/System.Linq.Dynamic.Core.Tests/DynamicExpressionParserTests.cs#L60

- For code-factor, see my post above. I'll check why you need to login to see these comments...

test/System.Linq.Dynamic.Core.Tests/DynamicExpressionParserTests.cs
```diff
@@ -452,5 +454,56 @@ public void ParseLambda_With_InnerStringLiteral()
...
+        {
+            Guid someId = Guid.NewGuid();
+            Guid anotherId = Guid.NewGuid();
+            var user = new User();
+            user.Id = someId;
+            Guid guidEmpty = Guid.Empty;
+            string expressionText = $"iif(@0.Id == \"{someId}\", Guid.Parse(\"{guidEmpty}\"), Guid.Parse(\"{anotherId}\"))";
+
+            var lambda = DynamicExpressionParser.ParseLambda(typeof(User), null, expressionText, user);
+
```
StefH: Why not also check here on:
``` c#
var boolLambda = lambda as Expression<Func<User, bool>>;
Assert.NotNull(boolLambda);
```
?

OlegNadymov: Yes. I've added it:
```
var guidLambda = lambda as Expression<Func<User, Guid>>;
Assert.NotNull(guidLambda);
```

OlegNadymov:
Ok! Now I got that you mean about // Arrange , // Act and // Assert.
I've added these comments. But I'm not sure about something like this:
```
            // Act
            var lambda = DynamicExpressionParser.ParseLambda(typeof(User), null, expressionText, user);
            var guidLambda = lambda as Expression<Func<User, Guid>>;
            Assert.NotNull(guidLambda);

            var del = lambda.Compile();
            Guid result = (Guid) del.DynamicInvoke(user);

            // Assert
            Assert.Equal(guidEmpty, result);
```
As you can see here the first ``Assert.NotNull`` is located inside Act. Is it ok?

StefH:
@OlegNadymov No need to mention me in the comments.

StefH:
> As you can see here the first Assert.NotNull is located inside Act. Is it ok?

for now it's ok.
What I sometimes do is use an `// Act 1` and `// Assert 1` and `// Act 2` and `// Assert 2`
Or just move all acts and asserts together.

OlegNadymov:
Ok! I'll do it like you said with several Act1, Act2 in the next PR.

OlegNadymov:
@StefH, what about this PR?

StefH:
I will merge this weekend and create new nuget.

OlegNadymov:
Ok! Then I'll add one more change for this PR.

OlegNadymov:
And one more change for LessGreater parsing.

StefH:
Maybe it's best to revert the last changes you did for LessGreater (unit test fail?). And create a new PR for this.

OlegNadymov:
I've fixed unit test. It was problem with custom static type

OlegNadymov:
Do you want that I create new PR? Or I can leave it here. In case with new PR, I need to create new branch for new PR because the current branch is using for this open PR. Do I understand correctly?

StefH:
OK, keep in same branch. I'll check the new code in some time...

OlegNadymov:
OK! Thank you! I'm looking forward to reviewing my changes. :)

src/System.Linq.Dynamic.Core/Parser/SupportedOperands/IEqualitySignatures.cs
```diff
@@ -13,9 +13,13 @@ internal interface IEqualitySignatures : IRelationalSignatures
 
         void F(Guid x, Guid y);
         void F(Guid? x, Guid? y);
-        void F(Guid x, string y);
-        void F(Guid? x, string y);
-        void F(string x, Guid y);
-        void F(string x, Guid? y);
+
+        // Disabled 2 lines below according to unit test ParseLambda_With_Guid_Equals_Null
```
StefH: Change comment to:

``` c#
// Disabled 4 lines below because of : https://github.com/StefH/System.Linq.Dynamic.Core/pull/200
//void F(Guid x, string y);
//void F(Guid? x, string y);
//void F(string x, Guid y);
//void F(string x, Guid? y);
```

test/System.Linq.Dynamic.Core.Tests/DynamicExpressionParserTests.cs
```diff
@@ -452,5 +504,146 @@ public void ParseLambda_With_InnerStringLiteral()
...
+        }
+
+        [Fact]
+        public void ParseLambda_With_Concat_String_CustomType()
+        {
+            // Arrange
+            string name = "name1";
+            string note = "note1";
+            var textHolder = new TextHolder(name, note);
+            string expressionText = $"Name + \" (\" + Note + \")\"";
```
StefH: No need to use $ here I think ? It's just a normal string.

test/System.Linq.Dynamic.Core.Tests/DynamicExpressionParserTests.cs
```diff
@@ -452,5 +504,146 @@ public void ParseLambda_With_InnerStringLiteral()
...
+        }
+
+        [Fact]
+        public void ParseLambda_With_Concat_CustomType_String()
+        {
+            // Arrange
+            string name = "name1";
+            string note = "note1";
+            var textHolder = new TextHolder(name, note);
+            string expressionText = $"Note + \" (\" + Name + \")\"";
```
StefH: No need to use $ here I think ? It's just a normal string.

test/System.Linq.Dynamic.Core.Tests/DynamicExpressionParserTests.cs
```diff
@@ -452,5 +504,146 @@ public void ParseLambda_With_InnerStringLiteral()
...
+            // Act 2
+            var del = lambda.Compile();
+            string result = (string) del.DynamicInvoke(textHolder);
+
+            // Assert 2
+            Assert.Equal("note1 (name1)", result);
+        }
+
+        [Fact]
+        public void ParseLambda_With_Less_Greater()
```
StefH: rename to `ParseLambda_Operator_Less_Greater_With_Guids` ?

test/System.Linq.Dynamic.Core.Tests/DynamicExpressionParserTests.cs
```diff
@@ -452,5 +504,146 @@ public void ParseLambda_With_InnerStringLiteral()
...
+                CustomTypeProvider = new TestCustomTypeProvider()
+            };
+
+            // Arrange
+            Guid someId = Guid.NewGuid();
+            Guid anotherId = Guid.NewGuid();
+            var user = new User();
+            user.Id = someId;
+            Guid guidEmpty = Guid.Empty;
+            string expressionText = $"iif(@0.Id == StaticHelper.GetGuid(\"name\"), Guid.Parse(\"{guidEmpty}\"), Guid.Parse(\"{anotherId}\"))";
```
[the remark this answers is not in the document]
OlegNadymov: This expression with ``Guid constant`` instead of ``Guid method`` was worked correctly before these changes. Therefore for represent bug with method I've introduced ``StaticHelper``. Or do you mean why I didn't create this class like as local class?

OlegNadymov:
All review's changes was done.

test/System.Linq.Dynamic.Core.Tests/DynamicExpressionParserTests.cs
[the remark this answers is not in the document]
StefH: I thought that the <> was not working at all. But I understand from your comment that this was only not working in specific flows?

In that case your test looks good. 

Maybe you could add some more comments to explain this?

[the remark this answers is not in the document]
OlegNadymov: The ``<>`` was working with constant only.
Your ``ExpressionParser`` has this condition (line 494):
```
else if ((constantExpr = right as ConstantExpression) != null && constantExpr.Value is string && (typeConverter = TypeConverterFactory.GetConverter(left.Type)) != null)
{
    right = Expression.Constant(typeConverter.ConvertFromInvariantString((string)constantExpr.Value), left.Type);
}
```

Therefore if ``right`` epxression is ``ConstantExpression`` then the <> will be working.
But for non-constant and for different types (e.g. ``Guid`` and ``Guid?``) will be executed this code (line 531):
```
CheckAndPromoteOperands(isEquality ? typeof(IEqualitySignatures) : typeof(IRelationalSignatures), op.Text, ref left, ref right, op.Pos);
```
And before my changes the variable ``isEquality`` was ``false``. Therefore into method ``CheckAndPromoteOperands`` passed argument type as ``typeof(IRelationalSignatures)``. But type ``typeof(IRelationalSignatures)`` doesn't have conversion between ``Guid`` and ``Guid?``.


Moreover for the ``!=`` this case was working because for the ``!=`` the variable ``isEquality``  was true:
```
bool isEquality = op.Id == TokenId.Equal || op.Id == TokenId.DoubleEqual || op.Id == TokenId.ExclamationEqual;
```
``TokenId.ExclamationEqual`` is equal to the ``!=``.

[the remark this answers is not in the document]
StefH: OK, thanks for the detailed explanation.

OlegNadymov:
Thank you for merge! And please nuget.

StefH:
1.0.8.17 is added to NuGet

OlegNadymov:
@StefH Thanks!

PR #201, opened 2018-09-04
Fix Parsing Config not passed down to expression parser in JOIN

david-garcia-garcia:
The ParsingConfiguration object is not properly propagated to CheckOuterAndInnerTypes() when calling the extension method Join.


david-garcia-garcia:
I checked usages of ParseLambda() and all 27 references are properly propagating a ParsingConfig after the change.

There is an overload that has 0 references, and which does not take a ParsingConfig.

StefH:
Can you point me to that overload which has 0 references?

PR #207, opened 2018-10-16
Fix New() support  for Type + Fix GroupJoin() not working when using Linq-To-Entities

### LINKED ISSUE #204: Cannot GroupJoin when source is Linq-To-Entities
I am struggling to make an outter join using this library.

After some digging I found that GroupJoin is passing the inner collection as a Constant Expression, where it should honor the original one:

 ``` c#
src/System.Linq.Dynamic.Core/DynamicQueryableExtensions.cs | 2 +-
 1 file changed, 1 insertion(+), 1 deletion(-)

diff --git a/src/System.Linq.Dynamic.Core/DynamicQueryableExtensions.cs b/src/System.Linq.Dynamic.Core/DynamicQueryableExtensions.cs
index 0a32c0d..7a487fe 100644
--- a/src/System.Linq.Dynamic.Core/DynamicQueryableExtensions.cs
+++ b/src/System.Linq.Dynamic.Core/DynamicQueryableExtensions.cs
@@ -697,7 +697,7 @@ namespace System.Linq.Dynamic.Core
                 typeof(Queryable), nameof(Queryable.GroupJoin),
                 new[] { outer.ElementType, innerType, outerSelectorLambda.Body.Type, resultSelectorLambda.Body.Type },
                 outer.Expression,
-                Expression.Constant(inner),
+                inner.AsQueryable().Expression,
                 Expression.Quote(outerSelectorLambda),
                 Expression.Quote(innerSelectorLambda),
                 Expression.Quote(resultSelectorLambda)));
```
I can see other extension methods such as Join use inner.AsQueryable().Expression instead of Expressoin.Constant(...).

When using Expression.Constant() EF will complain that it cannot materialize the object.

### LINKED ISSUE #206: Issue: new() expression cannot handle complex types
Currently, the new() expression can be used to instantiate know types, such as:

`query = query.select("new MyNamespace.MyClass(value1 as prop1, value2 as prop2)");`

ExpressionParser.cs is not able to properly handle complex types, such as generics or classes nested inside other classes, i.e.

`query = query.select("new MyNamespace.MyClass+MySubClass(value1 as prop1, value2 as prop2)");`

The parser is properly extracting the full namespace and class name, it just fails to properly detect the type in it's FindType() method. I really don't understand why the implementation of such a method is so convoluted, but following the current philosophy of the method, this is the patch that will allow new() to parse nested class names:

image

To properly support generics though some extra work needs to be done, as it will only consider dot and plus as valid tokens in a class name:

`while (_textParser.CurrentToken.Id == TokenId.Dot || _textParser.CurrentToken.Id == TokenId.Plus)`

### PR DESCRIPTION
**Pull request by david-garcia-garcia**
Fixes #204 and partially fixes #206.

### DIFF
--- a/test/System.Linq.Dynamic.Core.Tests/QueryableTests.Select.cs
+++ b/test/System.Linq.Dynamic.Core.Tests/QueryableTests.Select.cs
@@ -1,6 +1,5 @@
 ﻿using System.Collections;
 using System.Collections.Generic;
-using System.Data.Entity.Infrastructure;
 using System.Linq.Dynamic.Core.Exceptions;
 using System.Linq.Dynamic.Core.Tests.Helpers.Models;
 using Linq.PropertyTranslator.Core;
@@ -241,10 +240,13 @@ namespace System.Linq.Dynamic.Core.Tests
         [Fact]
         public void Select_Dynamic_IntoKnownNestedType()
         {
+            var config = new ParsingConfig();
+            config.AllowNewToEvaluateAnyType = true;
+
             // Act
             IQueryable<string> data = new List<string>() { "name1", "name2" }.AsQueryable();
             IQueryable<Example.NestedDto> projectedData = 
-                (IQueryable<Example.NestedDto>) data.Select($"new {typeof(Example.NestedDto).FullName}(~ as Name)");
+                (IQueryable<Example.NestedDto>) data.Select(config, $"new {typeof(Example.NestedDto).FullName}(~ as Name)");
 
             // Assert
             Check.That(projectedData.First().Name).Equals("name1");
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: test/System.Linq.Dynamic.Core.Tests/QueryableTests.Select.cs [hunk 1]
```diff
@@ -1,5 +1,6 @@
 ﻿using System.Collections;
 using System.Collections.Generic;
+using System.Data.Entity.Infrastructure;
```
StefH: This namespace is not available in the dotnet core version from the tests, so this fails.

david-garcia-garcia:
I had some tough time making tests work in my local environment.

So here is the deal with New() supporting loading any Type:

* FindType() in expression parsers was originally designed to ONLY allow loading types that appeard in the parsing context (it, root, parent), I guess this is due to SECURITY reasons. That is why I added an extra setting in ParsingConfig to allow loading any Type.
* Original tests where failing because, by default, Type.GetType() can only load types in the current assembly, so I had to add a more flexible resolver in the ParsingConfig iself.

PR #208, opened 2018-10-18
Fix New() support for Type + Fix GroupJoin() not working when using Linq-To-Entities (2)

StefH:
based on https://github.com/StefH/System.Linq.Dynamic.Core/pull/207


StefH:
@david-garcia-garcia ; can you take a look at my changes in this PR?

david-garcia-garcia:
I made a small comment. Overall looks good, an a much better approach to controlling the type resolvers.

Just a couple of comments related to this PR:

* This will be a breaking change, as IDynamicLinkCustomTypeProvider has grown. It is very very easy to fix, and probably IDynamicLinkCustomTypeProvider is not used by most implementation of this library.
* An indirect consequence of the new capabalities of the library is that the PromoteExpression method in Expression parser needs some extra love because when destination DTO is strongly typed then type matching is way more strict and this method needs to be more powerful to deal with int32-int64, Guid's and the like, I already have some changes in that respect, but it's better for a follow-up

StefH:
1. I will update version to the next one. So 1.0.9.0

2. Ok lets first complete this PR and make sure it works perfect.

Thanks for your help.

PR #229, opened 2018-11-25
Override is linq to objects

david-garcia-garcia:
Current implementation of IsLinqToObjects might no always return the correct result because there is no specificatin or reliable way of actually answering the question "IsLinqToObjects".

I..e in our project we are using LinqKit's AsExpandable() and we had to extend the current implementation by forking:

``` c#
private static bool IsProviderEnumerableQuery(IQueryProvider provider, IQueryable query)
        {
            Type baseType = provider.GetType().GetTypeInfo().BaseType;

            if (baseType == typeof(EnumerableQuery))
            {
                return true;
            }

            // Support for QueryTranslatorProvider
            if (baseType.Name == "QueryTranslatorProvider")
            {
                try
                {
                    IQueryProvider provider1 = baseType.GetProperty("OriginalProvider").GetValue(provider, (object[])null) as IQueryProvider;
                    return provider1 != null && QueryExtensions.IsProviderEnumerableQuery(provider1, query);
                }
                catch
                {
                    return false;
                }
            }

            // Support for LinqKit's ExpandableQuery
            if (provider.GetType().Name.StartsWith("ExpandableQuery"))
            {
                try
                {
                    IQueryable innerQuery = query.GetType().GetProperty("InnerQuery", BindingFlags.NonPublic | BindingFlags.Instance).GetValue(query) as IQueryable;
                    return innerQuery != null && QueryExtensions.IsProviderEnumerableQuery(innerQuery.Provider, innerQuery);
                }
                catch
                {
                    return false;
                }
            }

            return false;
        }
```

The proposal is to move IsLinqToObjects to the ParsingConfig object, so that users can customize it depending on what libraries they are using.

This is of course a breaking change as it breaks the current's library public contract.


src/System.Linq.Dynamic.Core/ParsingConfig.cs
```diff
@@ -68,6 +70,48 @@ public IExpressionPromoter ExpressionPromoter
...
+                    {
+                        PropertyInfo property = baseType.GetProperty("OriginalProvider");
+                        IQueryProvider originalProvider = property.GetValue(provider, null) as IQueryProvider;
+                        return originalProvider != null && IsProviderEnumerableQuery(originalProvider);
+                    }
+                    catch
+                    {
+                        return false;
+                    }
+                }
```
StefH: You are allowed to also add the Linqkit if-statement code here.
Can you please do that ?

StefH:
Good point, thanks for the PR.

However I would propose this:


1. Create a new Interface and Default implementation class.
2. Move the logic from `IsProviderEnumerableQuery` to that class.
3. Create a new property in the `ParsingConfig` class and build this property like `public IDynamicLinkCustomTypeProvider CustomTypeProvider`

For item 1:
Interface name could be:
``` c#
public interface IQueryableProviderAnalyzer
{
    bool IsProviderEnumerableQuery(IQueryable source)
}
```

And default implementation is
``` c#
public class DefaultQueryableProviderAnalyzer
{
    public bool IsProviderEnumerableQuery(IQueryable source)
    {
      // copy code from ParsingConfig -> IsProviderEnumerableQuery method here
    }
}
```

StefH:
I will take a look and try to refactor the code...

StefH:
@david-garcia-garcia - I did update the code, please take a look.

StefH:
@david-garcia-garcia : can you please take a look at my updates for this PR?

david-garcia-garcia:
Looks impecable! Been bussy...

PR #264, opened 2019-04-17
Fix escape characters parsing

StefH:
https://github.com/StefH/System.Linq.Dynamic.Core/issues/163


StefH:
https://github.com/StefH/System.Linq.Dynamic.Core/issues/163

PR #285, opened 2019-08-09
Fix certain cases where implicit conversions aren't correctly detected when parsing comparison operators

alexweav:
Hello, all.

I discovered a few edge cases where implicit conversions defined between two types would not be used when parsing comparison operators:

- The parser only searches for implicit conversions on the type which is being converted, and not the target type. If the conversion is only defined on the target type, it is missed, resulting in a parse failure.
- The parser doesn't check for implicit conversions at all when parsing equality comparisons on non-value types. If both types are non-value types (e.g. class and string), implicit conversions won't be considered at all.

I added a test which reproduces both cases, as well as fixes for these issues. The test now passes with the fixes for both issues applied.

Please let me know how I can improve the fixes here. Thanks!

### DIFF
--- a/src/System.Linq.Dynamic.Core/Parser/ExpressionParser.cs
+++ b/src/System.Linq.Dynamic.Core/Parser/ExpressionParser.cs
@@ -543,9 +543,16 @@ namespace System.Linq.Dynamic.Core.Parser
 
         private bool HasImplicitConversion(Type baseType, Type targetType)
         {
-            var methods = baseType.GetMethods(BindingFlags.Public | BindingFlags.Static).ToList();
-            methods.AddRange(targetType.GetMethods(BindingFlags.Public | BindingFlags.Static));
-            return methods
+            var baseTypeHasConversion = baseType.GetMethods(BindingFlags.Public | BindingFlags.Static)
+                .Where(mi => mi.Name == "op_Implicit" && mi.ReturnType == targetType)
+                .Any(mi => mi.GetParameters().FirstOrDefault()?.ParameterType == baseType);
+
+            if (baseTypeHasConversion)
+            {
+                return true;
+            }
+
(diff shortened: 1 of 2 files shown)


src/System.Linq.Dynamic.Core/Parser/ExpressionParser.cs
```diff
@@ -543,7 +543,9 @@ Expression ParseComparisonOperator()
 
         private bool HasImplicitConversion(Type baseType, Type targetType)
```
StefH: Wouldn't it be better to totally mirror the logic in this method?

Like:
``` c#
private bool HasImplicitConversion(Type baseType, Type targetType)
{
	bool baseTypeMatches = baseType.GetMethods(BindingFlags.Public | BindingFlags.Static)
		.Where(mi => mi.Name == "op_Implicit" && mi.ReturnType == targetType)
		.Any(mi => mi.GetParameters().FirstOrDefault()?.ParameterType == baseType);
		
	if (baseTypeMatches)
	{
		return true;
	}

	return targetType.GetMethods(BindingFlags.Public | BindingFlags.Static)
		.Where(mi => mi.Name == "op_Implicit" && mi.ReturnType == baseType)
		.Any(mi => mi.GetParameters().FirstOrDefault()?.ParameterType == targetType);
}
```

alexweav: Sure, I can format it that way. The actual type comparisons will be the exact same though (not mirrored, just duplicated), because this method only tests where there is an implicit conversion from `baseType` to `targetType` and not the other way around. We aren't guaranteed that the implicit conversion goes both ways - reversing the comparisons might mean that that we find the wrong conversion method.

alexweav:
Hello @StefH, any updates on this PR?

PR #290, opened 2019-08-22
Added SumAsync

wertzui:
The SumAsync method was missing from the Entity Framework Core extension methods.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.DynamicLinq/EFDynamicQueryableExtensions.cs
+++ b/src/Microsoft.EntityFrameworkCore.DynamicLinq/EFDynamicQueryableExtensions.cs
@@ -661,6 +661,92 @@ namespace EntityFramework.DynamicLinq
         }
         #endregion SingleOrDefault
 
+        #region SumAsync
+        private static readonly MethodInfo _sum = GetMethod(nameof(Queryable.Sum));
+
+        /// <summary>
+        ///     Asynchronously computes the sum of a sequence of values.
+        /// </summary>
+        /// <remarks>
+        ///     Multiple active operations on the same context instance are not supported.  Use 'await' to ensure
+        ///     that any asynchronous operations have completed before calling another method on this context.
+        /// </remarks>
+        /// <param name="source">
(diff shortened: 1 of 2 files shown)


src/Microsoft.EntityFrameworkCore.DynamicLinq/EFDynamicQueryableExtensions.cs
```diff
@@ -661,6 +661,92 @@ public static Task<dynamic> SingleOrDefaultAsync([NotNull] this IQueryable sourc
         }
         #endregion SingleOrDefault
 
+        #region SumAsync
+        private static readonly MethodInfo _sum = GetMethod(nameof(Queryable.Sum));
```
StefH: This one is correct I think.

src/Microsoft.EntityFrameworkCore.DynamicLinq/EFDynamicQueryableExtensions.cs
```diff
@@ -661,6 +661,92 @@ public static Task<dynamic> SingleOrDefaultAsync([NotNull] this IQueryable sourc
...
+        [PublicAPI]
+        public static Task<int> SumAsync([NotNull] this IQueryable source, CancellationToken cancellationToken = default(CancellationToken))
+        {
+            Check.NotNull(source, nameof(source));
+            Check.NotNull(cancellationToken, nameof(cancellationToken));
+
+            return ExecuteAsync<int>(_sum, source, cancellationToken);
+        }
+
+        private static readonly MethodInfo _sumSelector = GetMethod(nameof(Queryable.Sum), 1);
```
StefH: This cannot work because multiple overloads exist.

wertzui: I changed the code so it will always get the correct Sum method based on the type of the query.

StefH:
@wertzui Can you maybe add a SumAsync exmple to https://github.com/StefH/System.Linq.Dynamic.Core/blob/master/src-console/ConsoleAppEF1.1/Program.cs ? In that way you can be sure that this works for NETCore 1.x

PR #354, opened 2020-03-05
Fix correctly type cast of nulls (and other constants)

rockResolve:
Add tests that check returned type
Also fixes casts e.g. "int?(5)" to return nullable int type

### DIFF
--- a/src/System.Linq.Dynamic.Core/Parser/ExpressionParser.cs
+++ b/src/System.Linq.Dynamic.Core/Parser/ExpressionParser.cs
@@ -1518,10 +1518,12 @@ namespace System.Linq.Dynamic.Core.Parser
             {
                 Expression[] args = shorthand ? new[] { ParseStringLiteral() } : ParseArgumentList();
 
-                // If only 1 argument, and the arg is ConstantExpression, just return the ConstantExpression
-                if (args.Length == 1 && args[0] is ConstantExpression)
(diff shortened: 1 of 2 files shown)


src/System.Linq.Dynamic.Core/Parser/ExpressionParser.cs
```diff
@@ -1518,10 +1518,12 @@ Expression ParseTypeAccess(Type type)
             {
                 Expression[] args = shorthand ? new[] { ParseStringLiteral() } : ParseArgumentList();
 
-                // If only 1 argument, and the arg is ConstantExpression, just return the ConstantExpression
```
StefH: Is this ok?
I think the code + comments should be changed to this logic?

- If only 1 argument and not-null and ConstantExpression -> return the ConstantExpression
- If only 1 argument and null and ConstantExpression -> return the conversion (Can't use constructor)

rockResolve: The first case will cause other problems when casting e.g.  "int?(5)" would return an int, not a nullable int.
I have added a test for int?(5) which would otherwise fail because it would return int(5).

Note I use GenerateConversion so if the constant type *does* match the expression type, it will quickly return the constant.

StefH: OK

PR #357, opened 2020-03-16
Prioritize property or field over the type / Fix find for static property or field

konzen:
When the type and property have the same name the parser takes the property instead of type


konzen:
Changed Title

Included tests

StefH:
@konzen 
I needed to introduce a new config setting for this. Else existing code breaks. (--> #662)

Issue #351, opened 2020-02-21
Case insensitive GroupBy()


kKen94:
Is it possible adding IEqualityComparer to GroupBy()?

StefH:
Technically this is possible, however, what problem will this solve? Can you provide an example / unit test?

kKen94:
`var distinctedRaw = itemsToGroup
                .AsQueryable()
                .GroupBy($"new({string.Join(",", keys)})", "it")
                .Select("it.First()")
                .ToDynamicList<T>();`

I use GroupBy() to group same values by dynamic keys and get unique values to remove duplicates.
But if I don't apply UpperCase(), "STRing"and "strING" for GroupBy are not the same and when I go to save data on database, the latter (obviously) crashes

StefH:
If you execute this query on a database (sql), then this does work fine.

In LinqPad:
``` c#
var city = x.AsQueryable().Where(t => t.Text == "NewYork");
city.Dump();

var distinctedRaw = x
  .AsQueryable().Where(t => t.Text == "NewYork")
  .GroupBy($"new({string.Join(",", keys)})", "it")
  .Select("it.FirstOrDefault()");
  distinctedRaw.Dump();
```
![image](https://user-images.githubusercontent.com/249938/75354467-4ba65580-58ad-11ea-93e8-c376470cfaca.png)

kKen94:
The problem is that i'm not querying db, i'm manipulating local array

StefH:
I see.
It's possible to fix this, however it requires a lot of code changes in the GroupBy methods.

kKen94:
... and a lot of time I suppose. This issue is not urgent, at the moment I lowercasing all. It's a bad workaround, I know, but otherwise I wouldn't know how to do it

JonathanMagnan:
Hello @kKen94 ,

The v1.1.2 has been released.

The GroupBy now supports the `IEqualityComparer` overload.

Best Regards,

Jon

---

**Performance Libraries**
`context.BulkInsert(list, options => options.BatchSize = 1000);  `
[Entity Framework Extensions](http://entityframework-extensions.net/) • [Entity Framework Classic](http://entityframework-classic.net/) • [Bulk Operations](http://bulk-operations.net/) •  [Dapper Plus](http://dapper-plus.net/)

**Runtime Evaluation**
`Eval.Execute("x + y", new {x = 1, y = 2}); // return 3`
[...]

PR #381, opened 2020-05-26
save

Lempireqc:
Fix for: https://github.com/zzzprojects/System.Linq.Dynamic.Core/issues/351

### DIFF
--- a/Z.Dynamic.Core.Lab/Program.cs
+++ b/Z.Dynamic.Core.Lab/Program.cs
@@ -6,7 +6,7 @@ namespace Z.Dynamic.Core.Lab
 	{
 		static void Main(string[] args)
 		{
-            Request_OrderBy_StringComparer.Execute();
+            Request_GroupBy_IEqualityComparer.Execute();
 		}
 	}
 }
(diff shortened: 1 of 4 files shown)


StefH:
Hello @Lempireqc  and @JonathanMagnan,

It may be a good idea to give a meaningful description to this pull-request (and also to #380). This helps for understanding what has been added to the project.

Also it can be useful to link a PR to the related issue, in that way tracking the PR and the issue is easier.

And maybe add/update unit-tests for this one and the previous? Or will this be done in another PR?

JonathanMagnan:
Hello @StefH ,

You are right. I was planned to give more information on release note but it should be in the pull as well. I will update it when I will release this weekend. We maybe try sometime to save too much time ;)

This one already has a unit test for both pull: https://github.com/zzzprojects/System.Linq.Dynamic.Core/pull/381/files#diff-b752cf67daca39a6e6a249cc10ad554f

StefH:
For 1]
Previously I did generate the release-notes automatically using [GitHubReleaseNotes.exe](https://github.com/StefH/GitHubReleaseNotes) , https://github.com/zzzprojects/System.Linq.Dynamic.Core/blob/master/GitHubReleaseNotes.txt
These release notes were based on the Tags, PullRequests and Issues.
And the PR's and Issues need to have the correct Label applied.

The file looks like  : https://github.com/zzzprojects/System.Linq.Dynamic.Core/blob/master/CHANGELOG.md

For 3]
I see indeed that you have a unit-test, however, I was not expecting to find that file with that name in that location.
It's easier to keep the unit-tests grouped in the same structure, so you can see in one quick overview what tests are there.

PR #390, opened 2020-06-15
Fixed loading "Microsoft.EntityFrameworkCore.DynamicLinq.DynamicFunctions"

StefH:
Because of the file-version fix: another fix is needed #390 else the DynamicFunctions class is not loaded anymore.


StefH:
Because of the file-version fix: another fix is needed #390 else the DynamicFunctions class is not loaded anymore.

Issue #331, opened 2019-12-11
UseParameterizedNamesInDynamicQuery not working when parameter is converted


SamWhitby:
Hi,

The setting UseParameterizedNamesInDynamicQuery is not being respected when the parser has to apply a convert.

For example, i have the class with a nullable double property
```csharp
internal class Person
{
    public double? FavouriteDecimalNumber { get; set; }
}
```
and i have the following expression that checks for equality against a specific number
```csharp
var lambda = DynamicExpressionParser.ParseLambda<Person, bool>(new ParsingConfig() { UseParameterizedNamesInDynamicQuery = true }, false, "FavouriteDecimalNumber = 44.8");
```

The expression that is generated does not warp the value, instead it uses a constant

![image](https://user-images.githubusercontent.com/3334161/70582557-bb4a6400-1c1f-11ea-8cf3-60660216014b.png)

Making the property on the class Person non nullable removes the convert from the generated expression and the value then gets wrappped/parameterized as seen below
![image](https://user-images.githubusercontent.com/3334161/70582732-47f52200-1c20-11ea-8fa7-e33b42490327.png)

The problem seems to be in [here](https://github.com/StefH/System.Linq.Dynamic.Core/blob/57ce8f71d70c2e5075d258b8b4809736d2d76eb1/src/System.Linq.Dynamic.Core/Parser/ConstantExpressionWrapper.cs#L13) where the code is only checking for the type of ConstantExpression, however in this case the expression comes through as a UnaryExpression.  

I'm assuming this can be fixed by calling ConstantExpressionWrapper.Wrap before applying the cast function (but this seem to occur in quite a few places) or enhancing the ConstantExpressionWrapper.Wrap function to handle UnaryExpression types.

StefH:
@SamWhitby 
Can you try version `System.Linq.Dynamic.Core.1.0.20-ci-12330.nupkg` from MyGet ?

SamWhitby:
Thanks for the quick response.
Unfortunately that does not fix my issue. I grabbed your branch and ran your test [here](https://github.com/StefH/System.Linq.Dynamic.Core/blob/7991751236b9f6c68373ff4f174ae76f8d5b9ce4/test/System.Linq.Dynamic.Core.Tests/DynamicExpressionParserTests.cs#L281) and its passing under that scenario, however if you add a double property to the person object and change the value you are comparing to something like 42.4 you can see a cast is applied to the generated query and the value is not parameterized

[...]

PR #399, opened 2020-07-10
Use parameterized names in dynamic query

ascott18:
Builds on https://github.com/zzzprojects/System.Linq.Dynamic.Core/issues/331 - added wrapping of constants as soon as they're parsed to avoid cases where they can slip through the cracks.

### DIFF
--- a/src/System.Linq.Dynamic.Core/Parser/ExpressionParser.cs
+++ b/src/System.Linq.Dynamic.Core/Parser/ExpressionParser.cs
@@ -737,6 +737,8 @@ namespace System.Linq.Dynamic.Core.Parser
         Expression ParsePrimary()
         {
             Expression expr = ParsePrimaryStart();
+            _expressionHelper.WrapConstantExpression(ref expr);
+
             while (true)
             {
                 if (_textParser.CurrentToken.Id == TokenId.Dot)
--- a/test/System.Linq.Dynamic.Core.Tests/DynamicExpressionParserTests.cs
+++ b/test/System.Linq.Dynamic.Core.Tests/DynamicExpressionParserTests.cs
@@ -9,6 +9,7 @@ using System.Runtime.CompilerServices;
 using FluentAssertions;
 using Xunit;
 using User = System.Linq.Dynamic.Core.Tests.Helpers.Models.User;
+using System.Linq.Dynamic.Core.Tests.TestHelpers;
 
 namespace System.Linq.Dynamic.Core.Tests
 {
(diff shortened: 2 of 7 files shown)


src/System.Linq.Dynamic.Core/Parser/ExpressionParser.cs
```diff
@@ -737,6 +737,8 @@ Expression ParseUnary()
         Expression ParsePrimary()
         {
             Expression expr = ParsePrimaryStart();
+            _expressionHelper.WrapConstantExpression(ref expr);
```
ascott18, line 740: @StefH This is the principal change here. I'll let you determine if this makes sense to fix the issue in this way or not.

test/System.Linq.Dynamic.Core.Tests/DynamicExpressionParserTests.cs
```diff
@@ -261,16 +281,49 @@ public void DynamicExpressionParser_ParseLambda_UseParameterizedNamesInDynamicQu
...
-            string value = propertyInfo.GetValue(wrappedObj) as string;
+            int value = (int) propertyInfo.GetValue(wrappedObj);
 
-            Check.That(value).IsEqualTo("x");
+            Check.That(value).IsEqualTo(42);
+        }
+
+        [Theory]
+        [InlineData("NullableIntValue", "42")]
+        [InlineData("NullableDoubleValue", "42.23")]
```
ascott18, line 301: Test case for the issue in https://github.com/zzzprojects/System.Linq.Dynamic.Core/issues/331#issuecomment-564724355

ascott18:
@JonathanMagnan any updates on this? Would love to see this merged (or rejected, if that's the right answer).

JonathanMagnan:
Hello @ascott18 ,

Sorry for the delay,

I'm currently looking at it.

There is currently some weird stuff such as why a `double` constant require an `Expression.Convert` to be of type `Nullable<double>` while an `int` doesn't require such `Convert`. We already know where this is happening but we need to look deeper if such code is really required as you normally try to avoid Convert when not needed.

I'm not sure either if this line https://github.com/zzzprojects/System.Linq.Dynamic.Core/pull/399/files#diff-e02763e459f8622732aa03abaf297810R740 should really exists in the `ParsePrimary` or not.

I will continue to look at it and keep you updated later this week.

JonathanMagnan:
Hello @ascott18 ,

Your pull has been merged in the v1.2.1

Thank a lot for your help and your follow-up ;)

Best Regards,

Jon

PR #402, opened 2020-07-27
Ef core 3x support

JonathanMagnan:


StefH:
@JonathanMagnan 

I would not expect these duplicate files in the EF3 folder:
image

Was it not possible to use file links?

JonathanMagnan:
They are here only once in the EF Core 3 project: https://github.com/zzzprojects/System.Linq.Dynamic.Core/tree/master/src/Microsoft.EntityFrameworkCore.DynamicLinq.EFCore3

Other files were removed

StefH:
Perfect ! I did not expect these here.

I'm used to build it the other way : keep the files in the old and make links in the new.

But your solution also works perfectly !

JonathanMagnan:
Would have probably done it as well the way you initially did but it was too late so I choose to keep it like it ;)

Issue #404, opened 2020-08-06
Add support to Max and Min functions


DelvisR:
How to implement Max and Min aggregate functions:

public static object Max(this IQueryable source, string selector, params object[] values)

public static object Min(this IQueryable source, string selector, params object[] values)

JonathanMagnan:
Hello @DelvisR ,

If we try to support it, it will be without the `params object[] values` parameter.

Will it still be good for you?

Best Regards,

Jon

---

**Performance Libraries**
`context.BulkInsert(list, options => options.BatchSize = 1000);  `
[Entity Framework Extensions](http://entityframework-extensions.net/) • [Entity Framework Classic](http://entityframework-classic.net/) • [Bulk Operations](http://bulk-operations.net/) •  [Dapper Plus](http://dapper-plus.net/)

**Runtime Evaluation**
`Eval.Execute("x + y", new {x = 1, y = 2}); // return 3`
[C# Eval Function](http://eval-expression.net/) • [SQL Eval Function](http://eval-sql.net/)

JonathanMagnan:
Hello @DelvisR ,

As expected `Min` and `Max` are already supported: https://dotnetfiddle.net/mvH6yJ

One problem by adding your overload to `IQueryable`, is that some ORM such as `Entity Framework` will not be compatible with it. So it might be better to stick with the overload that `IQueryable` provide to avoid some future confusion.

DelvisR:
Even without parameters it would be very helpful.

Thank you,

DelvisR:
> 
> 
> Hello @DelvisR ,
> 
> As expected `Min` and `Max` are already supported: https://dotnetfiddle.net/mvH6yJ
> 
> One problem by adding your overload to `IQueryable`, is that some ORM such as `Entity Framework` will not be compatible with it. So it might be better to stick with the overload that `IQueryable` provide to avoid some future confusion.

Thank you very much!

PR #428, opened 2020-10-09
Add support for IQueryable.Min and .Max

gregfullman:
Min and Max extension methods for IQueryable were not available.

### DIFF
--- a/src/System.Linq.Dynamic.Core/DynamicQueryableExtensions.cs
+++ b/src/System.Linq.Dynamic.Core/DynamicQueryableExtensions.cs
@@ -1294,6 +1294,152 @@ namespace System.Linq.Dynamic.Core
         }
         #endregion LongCount
 
+        #region Max
+        private static readonly MethodInfo _max = GetMethod(nameof(Queryable.Max));
+
+        /// <summary>
+        /// Computes the max element of a sequence.
+        /// </summary>
+        /// <param name="source">A sequence of values to calculate find the max for.</param>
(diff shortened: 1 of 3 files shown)


JonathanMagnan:
Thank you for the pull.

We will look if there is a reason why `Min` and `Max` is still not here in the `DynamicQueryableExtensions` class.

We will look at your pull next week

Best Regards,

Jon

---

**Performance Libraries**
`context.BulkInsert(list, options => options.BatchSize = 1000);  `
[Entity Framework Extensions](http://entityframework-extensions.net/) • [Entity Framework Classic](http://entityframework-classic.net/) • [Bulk Operations](http://bulk-operations.net/) •  [Dapper Plus](http://dapper-plus.net/)

**Runtime Evaluation**
`Eval.Execute("x + y", new {x = 1, y = 2}); // return 3`
[C# Eval Function](http://eval-expression.net/) • [SQL Eval Function](http://eval-sql.net/)

StefH:
This was addressed before #404

gregfullman:
@StefH I did notice #404 - as far as I could tell it specifically addressed the usage of Min/Max nested within a dynamic expression (e.g. the example ion the dotnetfiddle). It didn't account for the extension methods exposed for the IQueryable interface.

JonathanMagnan:
Hello @gregfullman ,

The v1.2.3 has been released. I also agree on this addresses a different problem.

@StefH if you believe otherwise, let me know what should be done instead

Issue #439, opened 2020-10-28
Question: DateTime constructor using ticks


zspitz:
The following:
```csharp
public class Person {
    public DateTime DOB { get; set; }
}
var selector = $"DOB > DateTime(633979008000000000)";
var prm = Parameter(typeof(Person));
var parser = new ExpressionParser(new[] { prm }, selector, new object[] { }, ParsingConfig.Default);
var expr = parser.Parse(null);
```
fails with:

> **System.Linq.Dynamic.Core.Exceptions.ParseException**: 'A value of type 'Int64' cannot be converted to type 'DateTime''

Presumably, because Dynamic LINQ interprets `DateTime(...)` as a conversion from the `long` literal value, which isn't found; before searching for the appropriate constructor.

Is this the intended behavior?

I can work around this by passing in `DateTimeKind` as the second argument to the constructor:
```csharp
var selector = "DOB > DateTime(633979008000000000, DateTimeKind.Unspecified)";
```
Is there any other way to force Dynamic LINQ to recognize this as a call to the constructor?

StefH:
Solved by https://github.com/zzzprojects/System.Linq.Dynamic.Core/pull/449

PR #449, opened 2020-11-05
Fix DateTime constructor using ticks

StefH:


StefH:
https://github.com/zzzprojects/System.Linq.Dynamic.Core/issues/439

Issue #448, opened 2020-11-02
Dynamic.DynamicIndex is exposed in the expression


hazzik:
After upgrading from 1.1.8 to 1.2.0 I'm getting MethodNotSupported exception because my expression parser finds `Dynamic.DynamicIndex(object, string)` method in the expression. Strangely this method was introduced in 1.1.8, but the error started happening only in 1.2.0

Here are the failing tests: https://ci.appveyor.com/project/hazzik/nhibernate-core/builds/35962348/job/uvi4bg3jew1giw41/tests

StefH:
Hello @hazzik, thanks for reporting.

I'll take a look.

StefH:
@hazzik 
Can you provide a full source code example which displays this error?

hazzik:
This is the simplest reproduction: https://dotnetfiddle.net/8Az1F6

It outputs
```text
Product[].Where(Param_0 => (DynamicIndex(Param_0.Properties, "Name") == Convert("First Product", Object)))
```
And DynamicIndex method call is a problem here.

StefH:
Hello @hazzik,
I think I found a possible fix for this issue, however I want to test this using NHibernate.
Can you please provide me with a minimal console-app which can reproduce this issue?

StefH:
Dear @hazzik , can you please provide a minimal console-app to reproduce this issue?

hazzik:
I'll do it on or by this weekend.

hazzik:
@StefH here is the console app with reproduction: https://github.com/hazzik/nhibernate-linq-dynamic

StefH:
@hazzik Thank you very much, I'll copy this example to verify/reproduce the bug.

StefH:
I'm working on a fix, but `dynamic` cannot be supported.

This is also not supported in normal code:
![image](https://user-images.githubusercontent.com/249938/106781725-a7469a80-6649-11eb-8f14-7d7739117b07.png)


The code is reworked that this scenario can work:
``` c#
class Product
	{
		public virtual Guid Id { get; set; }
		public virtual string Name { get; set; }
		public virtual IDictionary<string, object> Properties { get; set; }   // <--
	}
```

with:
``` c#
var properties1 = new Dictionary<string, object>
{
	["Name"] = "First Product",
	["Description"] = "First Description"
};
session.Save(
	new Product
	{
		Id = Guid.NewGuid(),
		Properties = properties1
	});

//var properties2 = new
[...]

Issue #397, opened 2020-07-06
Dictionary parameter issue


b0bi79:
If pass Dictionary as a parameter for a lambda expression, then DynamicInvoke will return the first result each time. It looks like a cached result is returning.

```c#
        [Fact]
        public void EvalDictionaryParams2()
        {
            object CreateDicParameter(string name) => new Dictionary<string, object>
                {{"Name", new Dictionary<string, object> {{"FirstName", name }}}};

            var config = new ParsingConfig()
            {
                CustomTypeProvider = new DefaultDynamicLinqCustomTypeProvider()
            };
            var parType = new Dictionary<string, object>().GetType();
            var lambda = DynamicExpressionParser.ParseLambda(config, new [] {Expression.Parameter(parType, "item")}, typeof(object), "item.Name.FirstName").Compile();
            lambda.DynamicInvoke(CreateDicParameter("Julio")).Should().Be("Julio");
            lambda.DynamicInvoke(CreateDicParameter("John")).Should().Be("John");  // Failure. DynamicInvoke returns "Julio"
        }
```

JonathanMagnan:
Hello @b0bi79 ,

Thank you for reporting.

We will look at it.

Best Regards,

Jon

---

**Performance Libraries**
`context.BulkInsert(list, options => options.BatchSize = 1000);  `
[Entity Framework Extensions](http://entityframework-extensions.net/) • [Entity Framework Classic](http://entityframework-classic.net/) • [Bulk Operations](http://bulk-operations.net/) •  [Dapper Plus](http://dapper-plus.net/)

**Runtime Evaluation**
`Eval.Execute("x + y", new {x = 1, y = 2}); // return 3`
[C# Eval Function](http://eval-expression.net/) • [SQL Eval Function](http://eval-sql.net/)

JonathanMagnan:
Hello @b0bi79 ,

Just to let you know we fixed the issue. We will complete our test tomorrow and probably release it during the weekend.

Best Regards,

Jon

JonathanMagnan:
Hello @b0bi79 ,

The v1.1.8 has been released.

Let us know if everything is now working as expected.

Best Regards,

Jon

PR #480, opened 2021-02-02
Fix DynamicIndex implementation

StefH:


src/System.Linq.Dynamic.Core/System.Linq.Dynamic.Core.csproj
```diff
@@ -77,6 +77,7 @@
     <ItemGroup Condition=" '$(TargetFramework)' == 'netstandard1.3' ">
         <PackageReference Include="System.ComponentModel.TypeConverter" Version="4.3.0" />
         <PackageReference Include="System.Dynamic.Runtime" Version="4.3.0" />
+        <PackageReference Include="Microsoft.CSharp" Version="4.3.0" />
```
hazzik, line 80: Do you actually need this reference?

hazzik: No, it is not needed after f907489ce1c8818f5466a87bdc5377c8938ba5a7 where you exclude `DynamicGetMemberBinder` from netstandard1.3

StefH:
Hello @JonathanMagnan,

Can you or a developer please take a look at this PR ? 

This PR solves issue #448 but in order to fix that I had to remove the **DynamicIndex** logic which was introduced in 1.1.8 and use the original implementation (with a small change) to make sure that #397 keeps working correctly.

For that, I did enable the unittest again:
https://github.com/zzzprojects/System.Linq.Dynamic.Core/pull/480/files#diff-8c64d2f034f0d47405c4723ad04680f885cd71c73a5b0f498d04e95949789348L95

Please review the changes.

src/System.Linq.Dynamic.Core/System.Linq.Dynamic.Core.csproj
StefH: Thanks. It's removed.

hazzik:
@StefH with following implementation I was able to get a robust solution for my problem as well as un-comment `ExpressionTests_Select_DynamicObjects` test

```
internal class DynamicGetMemberBinder : GetMemberBinder
{
    private static readonly Type DynamicWrapperType = typeof(DynamicWrapper);
    private static readonly ConstructorInfo DynamicWrapperConstructor = DynamicWrapperType.GetConstructor(new[] {typeof(object)});
    private static readonly PropertyInfo DynamicWrapperTypeIndexer = DynamicWrapperType.GetProperty("Item");

    public DynamicGetMemberBinder(string name, [CanBeNull] ParsingConfig config) : base(name, !(config?.IsCaseSensitive == true))
    {
    }

    public override DynamicMetaObject FallbackGetMember(DynamicMetaObject target, DynamicMetaObject errorSuggestion)
    {
        var instance = Expression.New(DynamicWrapperConstructor, target.Expression);
        var indexExpression = Expression.MakeIndex(instance, DynamicWrapperTypeIndexer, new Expression[] {Expression.Constant(Name)});
        return DynamicMetaObject.Create(target.Value, indexExpression);
    }

    private class DynamicWrapper
    {
        private readonly object _item;
        private readonly Dictionary<string, PropertyInfo> _properties;

        public DynamicWrapper(object item)
        {
            _item = item;
            if (item != null && !(item is IDictionary<string, object>))
            {
                _properties = item.GetType().GetProperties()
                    .ToDictionary(p => p.Name, d => d);
            }
        }

        public object this[string value]
        {
            get
            {
                if (_item is null)
                    throw new NullReferenceException();
                
                if (_item is IDictionary<string, object> dictionary)
                    return dictionary[value];

                if (_properties.TryGetValue(value, out var property))
                    return property.GetValue(_item, null);

                throw new InvalidOperationException();
            }
        }
    }
}
```

JonathanMagnan:
Hello @StefH ,

The `PR` looks good to me.

PR #484, opened 2021-02-11
Implement support for anonymous types as dynamic objects

hazzik:
It also allows mixing different types of the dynamic objects in the same queryable.


test/System.Linq.Dynamic.Core.Tests/ExpressionTests.cs
```diff
@@ -1610,12 +1610,12 @@ public void ExpressionTests_OrderBy()
...
         }
 
-        //[Fact]
+        [Fact]
         public void ExpressionTests_Select_DynamicObjects()
         {
             // Arrange
             dynamic a1 = new { Name = "a", BlogId = 100 };
-            dynamic a2 = new { Name = "b", BlogId = 200 };
+            dynamic a2 = new { BlogId = 200, Name = "b" };
```
hazzik, line 1618: Properties here are switched intentionally to use different anonymous classes

StefH:
Thanks a lot. I'll merge this.

StefH (approved):

Issue #487, opened 2021-02-22
DynamicExpressionParser.ParseLambda can not parse 'TrimEnd' string method


Evgentus0:
Here is what to include in your request to make sure we implement a solution as quickly as possible. 

## 1. Description
Trying parse expression which using 'TrimEnd' string method I get an error "No applicable method 'TrimEnd' exists in type 'String''"

## 2. Exception
- Message : "No applicable method 'TrimEnd' exists in type 'String'"
- StackTrace: <br />
   at System.Linq.Dynamic.Core.Parser.ExpressionParser.ParseMemberAccess(Type type, Expression expression)
   at System.Linq.Dynamic.Core.Parser.ExpressionParser.ParseIdentifier()
   at System.Linq.Dynamic.Core.Parser.ExpressionParser.ParsePrimaryStart()
   at System.Linq.Dynamic.Core.Parser.ExpressionParser.ParsePrimary()
   at System.Linq.Dynamic.Core.Parser.ExpressionParser.ParseUnary()
   at System.Linq.Dynamic.Core.Parser.ExpressionParser.ParseMultiplicative()
   at System.Linq.Dynamic.Core.Parser.ExpressionParser.ParseAdditive()
   at System.Linq.Dynamic.Core.Parser.ExpressionParser.ParseShiftOperator()
   at System.Linq.Dynamic.Core.Parser.ExpressionParser.ParseComparisonOperator()
   at System.Linq.Dynamic.Core.Parser.ExpressionParser.ParseLogicalAndOrOperator()
   at System.Linq.Dynamic.Core.Parser.ExpressionParser.ParseIn()
   at System.Linq.Dynamic.Core.Parser.ExpressionParser.ParseAndOperator()
   at System.Linq.Dynamic.Core.Parser.ExpressionParser.ParseOrOperator()
   at System.Linq.Dynamic.Core.Parser.ExpressionParser.ParseLambdaOperator()
   at System.Linq.Dynamic.Core.Parser.ExpressionParser.ParseNullCoalescingOperator()
   at System.Linq.Dynamic.Core.Parser.ExpressionParser.ParseConditionalOperator()
   at System.Linq.Dynamic.Core.Parser.ExpressionParser.Parse(Type resultType, Boolean createParameterCtor)
[...]

PR #488, opened 2021-02-22
If args count is 0 -> parametereless method is better than method with parameters

AndriiZ:
Resolves issue https://github.com/zzzprojects/System.Linq.Dynamic.Core/issues/487


StefH:
https://github.com/zzzprojects/System.Linq.Dynamic.Core/issues/492

Issue #499, opened 2021-04-13
Feature: Support for Concat and optionally Union, Except, Intersect


jbhelm:
It does not appear there is support for the concatenation functions `Concat`, `Union`, `Except`, or `Intersect`. Would it be possible to add support for these? The most important one for my use case is `Concat`, while the others would just be nice-to-have to round out the functionality.

My use case would be to use the function inside a Select() string, for example:
`Select("CollectionA.Concat(CollectionB).Sum(SomeValue)")`

StefH:
@jbhelm 
See 1.2.10-preview-01 version at https://www.myget.org/feed/Packages/system-linq-dynamic-core

StefH:
Hello @jbhelm, did you have time to test this preview version?

jbhelm:
Hi @StefH  --  I've never MyGet. Following the link provided takes me to a login prompt , I do not have a MyGet account. I also tried adding the URL as a NuGet package source in Visual Studio, but that does not work either. I'm not sure what I'm supposed to do with that URL.

StefH:
Hello @jbhelm, see this link which describes how to use MyGet for NuGet packages:
https://github.com/zzzprojects/System.Linq.Dynamic.Core/wiki/MyGet-preview-versions

StefH:
A new NuGet version which includes this functionality will be released.

PR #506, opened 2021-05-09
Add Concat, Union, Except and Intersect

StefH:


StefH:
https://github.com/zzzprojects/System.Linq.Dynamic.Core/issues/499

PR #524, opened 2021-06-22
Parse in operator before comparisons

igitur:
Hi,

Inspired by SQL `where` clauses, I'm used to the `in` operator having precedence over others like `and` and `or`.

For example, I'm used to `MainCompanyId in (1, 2) and Name in ("A")` to be parsed as `(((x.MainCompanyId == 1) OrElse (x.MainCompanyId == 2)) AndAlso (x.Name == "A"))`. 

I think currently the parser attempts to parse it to something like `((((x.MainCompanyId == 1) OrElse (x.MainCompanyId == 2)) AndAlso x.Name) == "A")`, which is invalid syntax.

A current workaround is to add additional parentheses and write the clause as `(MainCompanyId in (1, 2)) and (Name in ("A"))`. This parses successfully, but it feels quite verbose to me.

This PR fixes the issue by first parsing for `in`s before comparisons. It was a simple fix and I'm almost surprised that it didn't break any existing tests (as far as I can see). I added a unit test for this PR's change.

Hope I didn't miss anything and that you agree with change in operator precedence in principle.

### DIFF
--- a/test/System.Linq.Dynamic.Core.Tests/Parser/ExpressionParserTests.cs
+++ b/test/System.Linq.Dynamic.Core.Tests/Parser/ExpressionParserTests.cs
@@ -93,6 +93,20 @@ namespace System.Linq.Dynamic.Core.Tests.Parser
             Check.That(parsedExpression).Equals(result);
         }
 
+        [Fact]
+        public void Parse_ParseMultipleInOperators()
+        {
+            // Arrange
+            ParameterExpression[] parameters = { ParameterExpressionHelper.CreateParameterExpression(typeof(Company), "x") };
+            var sut = new ExpressionParser(parameters, "MainCompanyId in (1, 2) and Name in (\"A\", \"B\")", null, null);
+
+            // Act
+            var parsedExpression = sut.Parse(null).ToString();
+
+            // Assert
+            Check.That(parsedExpression).Equals("(((x.MainCompanyId == 1) OrElse (x.MainCompanyId == 2)) AndAlso ((x.Name == \"A\") OrElse (x.Name == \"B\")))");
+        }
+
         [Theory]
         [InlineData("string(\"\")", "")]
         [InlineData("string(\"a\")", "a")]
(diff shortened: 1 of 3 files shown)


StefH:
@igitur 
Thank you.

Can you also create a unit-test which really tests this functionality ? Maybe EntitiesTests ?

test/System.Linq.Dynamic.Core.Tests/Parser/ExpressionParserTests.cs
```diff
@@ -93,6 +93,20 @@ public void Parse_ParseAndOperator(string expression, string result)
             Check.That(parsedExpression).Equals(result);
         }
 
+        [Fact]
+        public void Parse_ParseMultipleInOperators()
```
igitur, line 97: @StefH Is this unit test not sufficient?

igitur: Or if you can direct me to another similar unit test that I can base a new test on, that will be great. I'm a bit overwhelmed by the entire codebase.

igitur:
> Can you also create a unit-test which really tests this functionality ? Maybe EntitiesTests ?

Done. Added new partial class in `EntitiesTests.In.cs`.

StefH (approved):
Thanks

igitur:
@StefH Any chance of a new release to include the changes?

StefH:
@igitur 
It should already be included:
image

igitur:
Oh, thanks!

Where do you see that?

I just kept looking at the Released entry on the repo's main page sidebar:
image

And then I assumed that my PR wasn't released yet.

StefH:
https://github.com/zzzprojects/System.Linq.Dynamic.Core/blob/master/CHANGELOG.md

StefH:
I have to find a way to automatically convert a Tag into a Release here in GitHub...

igitur:
> https://github.com/zzzprojects/System.Linq.Dynamic.Core/blob/master/CHANGELOG.md

Thanks. I wasn't monitoring that.

> I have to find a way to automatically convert a Tag into a Release here in GitHub...

Github already creates a draft release when you tag. To convert that to a final release isn't much extra effort. But I'm sure there will be some GitHub Actions script to do that.

StefH:
I found a github action which can do this:
https://github.com/softprops/action-gh-release

I've added this action on this project, so in the future, when I add a tag, this will be converted into a release.

igitur:
Great, I might use that in my own projects.

StefH:
In case you are interested in generating the release notes in the same I do: take a look at https://github.com/StefH/GitHubReleaseNotes

Issue #527, opened 2021-06-29
Do you have any plan to support interpolated String?


yangzhongke:
Like:
```csharp
var query = db.Customers
.Where($"City == {cityName} and Orders.Count >= {count}")
```
I'd like to contribute code to support it, can I pull a request?

StefH:
You can create a PR, make sure that you also add unit-tests.

yangzhongke:
I have created a PR. Thanks.

PR #530, opened 2021-07-01
Add support of Interpolated strings.

yangzhongke:
Add support of Interpolated strings. 
see [https://github.com/zzzprojects/System.Linq.Dynamic.Core/issues/527](https://github.com/zzzprojects/System.Linq.Dynamic.Core/issues/527)

I have added two test cases: QueryableTests.FormattableString.cs and EntitiesTests.FormattableString.cs. All the tests are passed:
image

### DIFF
--- a/src/System.Linq.Dynamic.Core/DynamicQueryableWithFormattableStringExtensions.cs
+++ b/src/System.Linq.Dynamic.Core/DynamicQueryableWithFormattableStringExtensions.cs
@@ -301,7 +301,7 @@ namespace System.Linq.Dynamic.Core
         public static IQueryable SelectManyInterpolated([NotNull] this IQueryable source, [NotNull] Type resultType, [NotNull] FormattableString selector)
         {
             string selectorStr = ParseFormattableString(selector, out object[] args);
-            return DynamicQueryableExtensions.SelectMany(source, resultType,selectorStr, args);
+            return DynamicQueryableExtensions.SelectMany(source, resultType, selectorStr, args);
         }
 
         public static IQueryable<TResult> SelectManyInterpolated<TResult>([NotNull] this IQueryable source, [NotNull] ParsingConfig config, [NotNull] FormattableString selector)
@@ -343,7 +343,7 @@ namespace System.Linq.Dynamic.Core
         public static IOrderedQueryable<TSource> OrderByInterpolated<TSource>([NotNull] this IQueryable<TSource> source, [NotNull] ParsingConfig config, [NotNull] FormattableString ordering, IComparer comparer)
         {
             string orderingStr = ParseFormattableString(ordering, out object[] args);
-            return DynamicQueryableExtensions.OrderBy(source, config, orderingStr, comparer,args);
+            return DynamicQueryableExtensions.OrderBy(source, config, orderingStr, comparer, args);
         }
 
         public static IOrderedQueryable<TSource> OrderByInterpolated<TSource>([NotNull] this IQueryable<TSource> source, [NotNull] ParsingConfig config, [NotNull] FormattableString ordering)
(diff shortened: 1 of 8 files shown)


StefH:
Hello @yangzhongke, thank you for your contribution.

However, I thought that this functionality would be integrated in the current code, and not creating new extension methods.

Can you change your PR?

yangzhongke:
If there are two overloaded methods as below:
1) Where(string predicate,params object[] args)
2)Where(FormattableString predicate)

Due to the [deliberate decision by the Microsoft](https://stackoverflow.com/questions/35770713/overloaded-string-methods-with-string-interpolation) , The code Where($"Id={id}") will invoke  Where(string predicate,params object[] args) instead of Where(FormattableString predicate),  which lead to my decision to create new method names, like WhereInterpolated instead of using Where.

StefH:
Can't you just add a new method like?

Existing...
``` c#
public static IQueryable Where([NotNull] this IQueryable source, [NotNull] ParsingConfig config, [NotNull] string predicate, params object[] args)
````

New...
``` c#
public static IQueryable Where([NotNull] this IQueryable source, [NotNull] ParsingConfig config, [NotNull] FormattableString predicate)
```

yangzhongke:
I can. But where($" id={id}")  will invoke the "string predicate" version instead of "FormattableString" version. I have mentioned the reason  before.

yangzhongke:
Even EF Core created two methods to execute sql statements, ExecuteSqlRaw(string version) and ExecuteSqlInterpolated(FormatableString version). So I reckon that the two different methods solution, Where and WhereInterpolated , is a tradeoff.

StefH (changes requested):

yangzhongke:
Agree with the renaming.
Should I do something to do the changes?

StefH:
@yangzhongke 
Yes, if possible, change your PR code.

yangzhongke:
I have renamed those two files and committed the changes.
There are two files: DynamicQueryableWithFormattableStringExtensionsand EFDynamicQueryableWithFormattableStringExtensions.
DynamicQueryableWithFormattableStringExtensions.cs contains: AllInterpolated and other XXXInterpolated methods, and EFDynamicQueryableWithFormattableStringExtensions contains: AllInterpolatedAsync and other XXXInterpolatedAsync methods.
DynamicQueryableWithFormattableStringExtensionsand .cs is already on the src/System.Linq.Dynamic.Core folder; because EFDynamicQueryableWithFormattableStringExtensions uses Microsoft.EntityFrameworkCore, just like EntityFrameworkDynamicQueryableExtensions.cs, so EFDynamicQueryableWithFormattableStringExtensions  should be on the folder Microsoft.EntityFrameworkCore.DynamicLinq.EFCore3.

src/System.Linq.Dynamic.Core/DynamicQueryableWithFormattableStringExtensions.cs
```diff
@@ -0,0 +1,425 @@
...
+            return DynamicQueryableExtensions.ThenBy(source,  orderingStr, comparer, args);
+        }
+
+        public static IOrderedQueryable<TSource> ThenByInterpolated<TSource>([NotNull] this IOrderedQueryable<TSource> source, [NotNull] FormattableString ordering)
+        {
+            string orderingStr = ParseFormattableString(ordering, out object[] args);
+            return DynamicQueryableExtensions.ThenBy(source, orderingStr, args);
+        }
+
+        private static string ParseFormattableString(FormattableString predicate,out object[] args)
```
StefH, line 415: Can you do a format-file to make sure all spaces / code is aligned correctly?

yangzhongke: Sorry, I have no idea of what you mentioned, all the indents keep consistent. Maybe you can accept this PR, and then format the codes according to your preference.

yangzhongke: If what should I do is to format the code using the ' Format Document' functionality  in Visual Studio ,  I could do it.

StefH: Indeed "Format Document", that's enough.

yangzhongke: I have formatted the code and pushed. Please check.

Issue #531, opened 2021-07-09
Number parsing rejecting doubles in scientific notation


alexweav:
## 1. Description
Hello, I ran into an issue with parsing doubles using scientific notation. I believe I have tracked the issue down, and I am hoping to provide a fix.

Given a class defining a `double` field:
``` c#
public class Test
{
    public double Value { get; set; }
}
```

The following Dynamic Linq expression results in a `ParseException` when passed to `DynamicExpressionParser.ParseLambda`:
```
Value < 1.2345E-4
```

where [this documentation](http://ak-dynamic-linq.azurewebsites.net/GettingStarted#literals) states that `1.2345E-4` is a valid double literal.

This appears to happen because `ExpressionParser.TryParseAsDouble` always passes `NumberStyles.Number` to `double.Parse` [here](https://github.com/zzzprojects/System.Linq.Dynamic.Core/blob/822d797e01dbad5964d1365f5774616a3e8f9562/src/System.Linq.Dynamic.Core/Parser/ExpressionParser.cs#L946). This flag specifically disallows exponents in doubles: https://docs.microsoft.com/en-us/dotnet/api/system.globalization.numberstyles?view=net-5.0

I noticed that `TryParseAsFloat` uses `NumberStyles.Float` instead, which _does_ allow exponents. However, parsing as float always fails in my example above, because the expression doesn't add a qualifier which forces it to be recognized as a float.

## 2. Exception
Here is the exception that is thrown by `ParseLambda`:

```
System.Linq.Dynamic.Core.Exceptions.ParseException
  HResult=0x80131500
  Message=Invalid real literal '2e2'
  Source=System.Linq.Dynamic.Core
  StackTrace:
   at System.Linq.Dynamic.Core.Parser.ExpressionParser.TryParseAsDouble(String text, Char qualifier)
   at System.Linq.Dynamic.Core.Parser.ExpressionParser.TryParseAsDecimal(String text, Char qualifier)
   at System.Linq.Dynamic.Core.Parser.ExpressionParser.TryParseAsFloat(String text, Char qualifier)
   at System.Linq.Dynamic.Core.Parser.ExpressionParser.ParseRealLiteral()
   at System.Linq.Dynamic.Core.Parser.ExpressionParser.ParsePrimaryStart()
   at System.Linq.Dynamic.Core.Parser.ExpressionParser.ParsePrimary()
[...]

PR #532, opened 2021-07-11
Fix NumberParsing for double with exponent

StefH:


src/System.Linq.Dynamic.Core/Parser/NumberParser.cs
```diff
@@ -29,73 +44,73 @@ public object ParseNumber(string text, Type type)
...
                 }
                 if (tp == typeof(float))
                 {
-                    return float.Parse(text, _config.NumberParseCulture);
+                    return float.Parse(text, _culture);
                 }
                 if (tp == typeof(double))
                 {
-                    return double.Parse(text, _config.NumberParseCulture);
+                    return double.Parse(text, _culture);
```
alexweav, line 109: Nice, I think the default will work great for us. FWIW, the default value seems to be `NumberStyles.Float | NumberStyles.AllowThousands`. According to the [somewhat confusing NumberStyles docs](https://docs.microsoft.com/en-us/dotnet/api/system.globalization.numberstyles?view=net-5.0#system-globalization-numberstyles-allowtrailingsign), this actually does exclude one flag that was enabled prior to this change, which was `NumberStyles.AllowTrailingSign`.

This flag is really strange, it lets me specify the sign after the number like in [this fiddle](https://dotnetfiddle.net/GbPFqn). I haven't seen any actual use cases of this in the wild, and I don't know if the parser even supports it. I definitely don't need trailing signs, just wanting to make sure that we aren't regressing this edge case.

test/System.Linq.Dynamic.Core.Tests/DynamicExpressionParserTests.cs
```diff
@@ -644,21 +644,114 @@ public void DynamicExpressionParser_ParseLambda_ParameterExpressionMethodCall_Re
...
+                new object[] { "de-DE", "-42d", -42d },
+                new object[] { "de-DE", "3,215d", 3.215d },
+                new object[] { "de-DE", "1,2345E-4", 0.00012345d },
+                new object[] { "de-DE", "1,2345E4", 12345d },
+
+                new object[] { null, "1D", 1d },
+                new object[] { null, "1d", 1d },
+                new object[] { null, "-42d", -42d },
+                new object[] { null, "3.215d", 3.215d },
+                new object[] { null, "1.2345E-4", 0.00012345d },
```
alexweav, line 732: These are looking great. We saw a few cases where there was a lowercase `e` in the exponent. `double.Parse` handles this with no issue, but it may be worth including a test for it.

src/System.Linq.Dynamic.Core/Parser/NumberParser.cs
alexweav: I just [tried it](https://dotnetfiddle.net/HHbzpm), and it looks like trailing signs are already rejected elsewhere in the parser. So, this isn't a regression at all, we can ignore the trailing sign flag 👍

test/System.Linq.Dynamic.Core.Tests/DynamicExpressionParserTests.cs
StefH: I did add this test.

src/System.Linq.Dynamic.Core/Parser/NumberParser.cs
StefH: OK

alexweav (approved):
Thanks again!

StefH:
https://github.com/zzzprojects/System.Linq.Dynamic.Core/issues/531

PR #557, opened 2021-11-24
bind expressions to correct constructor parameter

jonasdaniels:
``` c#
class test {
  int a;
  int b;
  int c; 
}
```
```
if ConstructorInfo ci.GetParameters() returns ParameterInfo[] {int a, int b, int c }
```

Picking them up by index works in most cases.
ex.
new(1 as a, 2 as b, 3 as c ) => {a = 1, b = 2, c = 3 }
but not if the identifiers are moved around(for example using the same type again)
new(1 as b, 2 as a, 3 as c ) => { a =1, b = 2, c = 3 } 

Proposed solution.
Make sure to map the correct expression based the identifier to the same(based on type and name) constructor parameter

### DIFF
--- a/test/System.Linq.Dynamic.Core.Tests/Parser/ExpressionParserTests.TypeAccess.cs
+++ b/test/System.Linq.Dynamic.Core.Tests/Parser/ExpressionParserTests.TypeAccess.cs
@@ -1,4 +1,5 @@
-﻿using System.Linq.Dynamic.Core.Parser;
+﻿using System.Collections.Generic;
+using System.Linq.Dynamic.Core.Parser;
 using System.Linq.Expressions;
 using FluentAssertions;
 using Xunit;
@@ -96,6 +97,29 @@ namespace System.Linq.Dynamic.Core.Tests.Parser
 
             // Assert
             expression.ToString().Should().Be("new Uri(\"https://www.example.com/\", Absolute)");
+        }
+        
+        [Theory]
+        [InlineData("new(2 as b, 1 as a)", "new(1 as a, 2 as b)")]
+        public void ParseTypeAccess_Via_Constructor_DynamicType_To_String(string newExpression, string newExpression2)
+        {
+            // Arrange
+            var parameter = Expression.Parameter(typeof(int));
+            var parameter2 = Expression.Parameter(typeof(int));
+            var returnType = DynamicClassFactory.CreateType(new List<DynamicProperty> {
+                new DynamicProperty("a", typeof(int)),
+                new DynamicProperty("b", typeof(int))
+            });
+
+            // Act
+            var parser = new ExpressionParser(new[] { parameter, parameter2 }, newExpression, new object[] { }, ParsingConfig.Default);
+            var parser2 = new ExpressionParser(new[] { parameter, parameter2 }, newExpression2, new object[] { }, ParsingConfig.Default);
+
+            var expression = parser.Parse(returnType);
+            var expression2 = parser2.Parse(returnType);
+            // Assert
+            expression.ToString().Should().Be("new "+ returnType.Name + "(a = 1, b = 2)");
+            expression2.ToString().Should().Be("new "+ returnType.Name + "(a = 1, b = 2)");
         }
     }
 }
--- a/test/System.Linq.Dynamic.Core.Tests/Parser/ExpressionParserTests.TypeAccess.cs
+++ b/test/System.Linq.Dynamic.Core.Tests/Parser/ExpressionParserTests.TypeAccess.cs
@@ -100,7 +100,8 @@ namespace System.Linq.Dynamic.Core.Tests.Parser
         }
         
         [Theory]
-        [InlineData("new(2 as b, 1 as a)", "new(1 as a, 2 as b)")]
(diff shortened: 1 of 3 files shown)


StefH:
@jonasdaniels 

Can you maybe add a unit test to:
https://github.com/zzzprojects/System.Linq.Dynamic.Core/blob/93d93d7b807d9881217a2841c3025ec4dc4aba69/test/System.Linq.Dynamic.Core.Tests/Parser/ExpressionParserTests.TypeAccess.cs ?

test/System.Linq.Dynamic.Core.Tests/Parser/ExpressionParserTests.TypeAccess.cs
```diff
@@ -97,5 +98,28 @@ public void ParseTypeAccess_Via_Constructor_String_And_UriKind_To_Uri()
             // Assert
             expression.ToString().Should().Be("new Uri(\"https://www.example.com/\", Absolute)");
         }
+        
+        [Theory]
+        [InlineData("new(2 as b, 1 as a)", "new(1 as a, 2 as b)")]
```
StefH, line 103: The InlineData should be used as:
```
[InlineData("test-value-1", "expected-value")]
```

So in this case the test should be like:
``` c#
[Theory]
[InlineData("new(2 as b, 1 as a)", "new XXX(a = 1, b = 2)")]
[InlineData("new(1 as a, 2 as b)", "new XXX(a = 1, b = 2)")]
public void ParseTypeAccess_Via_Constructor_DynamicType_To_String(string newExpression, string expectedExpression)
{
	// Arrange
	var parameter = Expression.Parameter(typeof(int));
	var parameter2 = Expression.Parameter(typeof(int));
	var returnType = DynamicClassFactory.CreateType(new List<DynamicProperty> {
		new DynamicProperty("a", typeof(int)),
		new DynamicProperty("b", typeof(int))
	});

	// Act
	var parser = new ExpressionParser(new[] { parameter, parameter2 }, newExpression, new object[] { }, ParsingConfig.Default);

	var expression = parser.Parse(returnType);

	// Assert
	expression.ToString().Should().Be(expectedExpression);
}
```


Note that the `XXX` should be the type you expect.

StefH (changes requested):
please take a look at my comments

StefH:
@jonasdaniels 

Can you double check if all tests run fine?

At my system, I get this test error:
```
﻿ System.Linq.Dynamic.Core.Tests.QueryableTests.Select_Dynamic_IntoTypeWithNullableProperties2
   Source: QueryableTests.Select.cs line 253
   Duration: 26 ms

  Message: 
System.InvalidOperationException : Sequence contains no matching element

  Stack Trace: 
ThrowHelper.ThrowNoMatchException()
Enumerable.First[TSource](IEnumerable`1 source, Func`2 predicate)
ExpressionParser.CreateNewExpression(List`1 properties, List`1 expressions, Type newType) line 1480
ExpressionParser.ParseNew() line 1408
ExpressionParser.ParseIdentifier() line 991
ExpressionParser.ParsePrimaryStart() line 772
ExpressionParser.ParsePrimary() line 741
ExpressionParser.ParseUnary() line 736
ExpressionParser.ParseMultiplicative() line 682
ExpressionParser.ParseAdditive() line 651
<5 more frames...>
ExpressionParser.ParseOrOperator() line 265
ExpressionParser.ParseLambdaOperator() line 245
ExpressionParser.ParseNullCoalescingOperator() line 232
ExpressionParser.ParseConditionalOperator() line 216
ExpressionParser.Parse(Type resultType, Boolean createParameterCtor) line 154
DynamicExpressionParser.ParseLambda(Type delegateType, ParsingConfig parsingConfig, Boolean createParameterCtor, ParameterExpression[] parameters, Type resultType, String expression, Object[] values) line 122
DynamicExpressionParser.ParseLambda(ParsingConfig parsingConfig, Boolean createParameterCtor, ParameterExpression[] parameters, Type resultType, String expression, Object[] values) line 97
DynamicExpressionParser.ParseLambda(ParsingConfig parsingConfig, Boolean createParameterCtor, Type itType, Type resultType, String expression, Object[] values) line 355
DynamicQueryableExtensions.Select[TResult](IQueryable source, ParsingConfig config, String selector, Object[] args) line 1803
QueryableTests.Select_Dynamic_IntoTypeWithNullableProperties2() line 264
```

StefH (changes requested):
I see that you have updated the unit-tests, but actually I was expecting that you would fix the code, because the unit-test did work before, and ideally the unit-test should continue working, also with your new code.

Sop can you please take a look at the code and see if you can fix the code so that the unit-test can stay the same?

jonasdaniels:
@StefH 
Reverted the test changes and added code to sequentially bind parameters. 

However it feels unintuitiv to bind to the field names instead of the parameters names. 

In the case:
QueryableTests.Select.cs line 265: "new (it as Time, DayOfWeek as DOWNull, DayOfWeek as DOW, Second as Sec, int?(Second) as SecNull)");" 
it might make more sence to bind towards the fields since we dont have contructor with these names?

StefH (approved):

PR #558, opened 2021-12-03
Allow assembly to be collected

hemirunner426:
Running this library from a collectable AssemblyLoadContext produces the following exception:

System.NotSupportedException: 'A non-collectible assembly may not reference a collectible assembly.'


StefH:
@hemirunner426 Thank you

PR #576, opened 2022-02-28
fix enums could not be cast to decimal

jogibear9988:


jogibear9988:
could you create a new nuget?

StefH:
I'll release it within a few days.

Issue #586, opened 2022-05-04
Parenthesis around an "In" expression raise an exception


cambirch:
Parenthesis around an "In" expression raise an exception `')' or operator expected (at index 16)` instead of working.

```
')' or operator expected (at index 16)
   at System.Linq.Dynamic.Core.Tokenizer.TextParser.ValidateToken(TokenId tokenId, String errorMessage) in C:\Dev\GitHub\System.Linq.Dynamic.Core\src\System.Linq.Dynamic.Core\Tokenizer\TextParser.cs:line 456
   at System.Linq.Dynamic.Core.Parser.ExpressionParser.ParseParenExpression() in C:\Dev\GitHub\System.Linq.Dynamic.Core\src\System.Linq.Dynamic.Core\Parser\ExpressionParser.cs:line 835
   at System.Linq.Dynamic.Core.Parser.ExpressionParser.ParsePrimaryStart() in C:\Dev\GitHub\System.Linq.Dynamic.Core\src\System.Linq.Dynamic.Core\Parser\ExpressionParser.cs:line 780
   at System.Linq.Dynamic.Core.Parser.ExpressionParser.ParsePrimary() in C:\Dev\GitHub\System.Linq.Dynamic.Core\src\System.Linq.Dynamic.Core\Parser\ExpressionParser.cs:line 741
   at System.Linq.Dynamic.Core.Parser.ExpressionParser.ParseUnary() in C:\Dev\GitHub\System.Linq.Dynamic.Core\src\System.Linq.Dynamic.Core\Parser\ExpressionParser.cs:line 736
   at System.Linq.Dynamic.Core.Parser.ExpressionParser.ParseMultiplicative() in C:\Dev\GitHub\System.Linq.Dynamic.Core\src\System.Linq.Dynamic.Core\Parser\ExpressionParser.cs:line 682
   at System.Linq.Dynamic.Core.Parser.ExpressionParser.ParseAdditive() in C:\Dev\GitHub\System.Linq.Dynamic.Core\src\System.Linq.Dynamic.Core\Parser\ExpressionParser.cs:line 651
   at System.Linq.Dynamic.Core.Parser.ExpressionParser.ParseShiftOperator() in C:\Dev\GitHub\System.Linq.Dynamic.Core\src\System.Linq.Dynamic.Core\Parser\ExpressionParser.cs:line 627
[...]

PR #587, opened 2022-05-04
Fixed: Parenthesis around an "In" expression raise an exception

cambirch:
Fixes: #586

Parser is greedy eating the next token inside an In expression when the expression is an Identifier.
This worked correctly when the expression is: `x in (1, 2)` but fails when `(x in @0)`.

Moved the token read into the Paren code branch which fixes the problem.


StefH:
@cambirch 
Thank you.

Issue #614, opened 2022-07-22
Join problem with inherited entities


Ibehem:
1. Description
When I try to join on a descendant of a class I got an error:
**x and y are incompatible**

While the same in Strongly typed Line works.

2. Fiddle or Project

Based on the pets and person project: 
https://dotnetfiddle.net/eEYaoU

Here is a sample:
I have a simple hierarchy:
``` c#
public class Samples
{
	public int Id { get; set; }
	public string Name { get; set; }
	public List<Result> Results { get; set; } 
}

public class SoilSamples : Samples
{
	public int Deep { get; set; }
}


public class Result
{
	public int Id { get; set; }
	public int Value { get; set; }
	public Samples Sample { get; set; }
}
```

StefH:
Hello @Ibehem,

Thanks for creating this issue.

I did find the issue and I've corrected it. Do you want to test a preview version?

Note that your join is not correct, it should be something like this:
``` c#
// Arrange
var magnus = new Person { Name = "Hedlund, Magnus" };
var terry = new Person { Name = "Adams, Terry" };
var charlotte = new Person { Name = "Weiss, Charlotte" };

var barley = new Pet { Name = "Barley", Owner = terry };
var boots = new Pet { Name = "Boots", Owner = terry };
var whiskers = new Pet { Name = "Whiskers", Owner = charlotte };
var daisy = new SpecialPet { Name = "Daisy", Owner = magnus, IsSpecial = true };

var people = new List<Person> { magnus, terry, charlotte };
var pets = new List<Pet> { barley, boots, whiskers, daisy };

// Act
var realQuery = people.AsQueryable()
	.Join(
		pets,
		person => person,
		pet => pet.Owner,
		(person, pet) => new { OwnerName = person.Name, Pet = pet.Name });
var realResult = realQuery.ToList();

var dynamicQuery = people.AsQueryable()
	.Join(
		pets,
		"it",
		"Owner",
		"new(outer.Name as OwnerName, inner.Name as Pet)");
var dynamicResult = dynamicQuery.ToDynamicList();

realResult.Should().BeEquivalentTo(dynamicResult);
```

EfrSyn:
Hello Stef, Yes I would love to test a preview version ! And thank you very much for solving this so quickly!

StefH:
A preview (`preview-02`) can be downloaded from MyGet:
https://www.myget.org/F/system-linq-dynamic-core/api/v3/index.json

[...]

PR #621, opened 2022-08-05
Fix Join on inherited class

StefH:


StefH:
https://github.com/zzzprojects/System.Linq.Dynamic.Core/issues/614

PR #630, opened 2022-09-12
Fix MethodFinder to return topmost implementation of the virtual method

### LINKED ISSUE #580: C# Expressions use Base Class Virtual Methodinfo
If you write a Expression where you call a Virtual Method, the C# Expression Parser uses the MethodInfo from the BaseClass where System.Dynamic.Linq uses the MethodInfo from the Derived Class.

see issue: https://github.com/linq2db/linq2db/issues/3475

jogibear9988:
uses the topmost implementation of the virtual method

### DIFF
--- a/src/System.Linq.Dynamic.Core/Parser/SupportedMethods/MethodFinder.cs
+++ b/src/System.Linq.Dynamic.Core/Parser/SupportedMethods/MethodFinder.cs
@@ -105,7 +105,7 @@ namespace System.Linq.Dynamic.Core.Parser.SupportedMethods
             if (applicable.Length == 1)
             {
                 MethodData md = applicable[0];
-                method = md.MethodBase;
+                method = ((MethodInfo)md.MethodBase).GetBaseDefinition();
                 args = md.Args;
             }
             else
--- a/test/System.Linq.Dynamic.Core.Tests/Parser/MethodFinderTest.cs
+++ b/test/System.Linq.Dynamic.Core.Tests/Parser/MethodFinderTest.cs
@@ -1,0 +1,30 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using System.Linq.Expressions;
+using static System.Console;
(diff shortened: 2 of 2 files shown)


test/System.Linq.Dynamic.Core.Tests/Parser/MethodFinderTest.cs
```diff
@@ -0,0 +1,30 @@
+﻿using System;
```
StefH, line 1: Do you need all these usings?

src/System.Linq.Dynamic.Core/Parser/SupportedMethods/MethodFinder.cs
```diff
@@ -105,7 +105,7 @@ public int FindBestMethodBasedOnArguments(IEnumerable<MethodBase> methods, ref E
             if (applicable.Length == 1)
             {
                 MethodData md = applicable[0];
-                method = md.MethodBase;
+                method = ((MethodInfo)md.MethodBase).GetBaseDefinition();
```
StefH, line 108: question:

Is md.MethodBase always a MethodInfo?

StefH (changes requested):

test/System.Linq.Dynamic.Core.Tests/Parser/MethodFinderTest.cs
jogibear9988: think so

src/System.Linq.Dynamic.Core/Parser/SupportedMethods/MethodFinder.cs
jogibear9988: if i read here: https://docs.microsoft.com/en-us/dotnet/api/system.reflection.methodinfo.getbasedefinition?view=net-6.0

yes

StefH (approved):

StefH:
@jogibear9988 

I merged to master, however I noticed that some unit tests fail.

I did fix it.

Issue #629, opened 2022-09-10
Error accessing values in DynamicClass - fails on second attempt


glittle:
This is broken in version 1.2.20.   Version 1.2.19 is working.

## 1. Description
With this statement:
````
 var result = DataContext.Users
              .Where(u => u.Id == id)
              .Select($"new {{{"Name,Email"}}}") 
              .ToDynamicList()
              .FirstOrDefault();
````
The `result` variable has two properties: Name and Email.

On first access to a property, it works.
On second access to any property it fails with the exception shown.

For example:
````
var a = result["Name"]; // okay
var b = result["Name"]; // crashes with exception
````

## 2. Exception
````
'result["Name"]' threw an exception of type 'System.ArgumentException'
    Data: {System.Collections.ListDictionaryInternal}
    HResult: -2147024809
    HelpLink: null
    InnerException: null
    Message: "An item with the same key has already been added."
    ParamName: null
    Source: "mscorlib"
    StackTrace: 
		"  at System.ThrowHelper.ThrowArgumentException(ExceptionResource resource)
		   at System.Collections.Generic.Dictionary`2.Insert(TKey key, TValue value, Boolean add)
		   at System.Linq.Dynamic.Core.DynamicClass.get_Properties() in C:\\Dev\\GitHub\\System.Linq.Dynamic.Core\\src\\System.Linq.Dynamic.Core\\DynamicClass.cs:line 36
		   at CallSite.Target(Closure , CallSite , Object , String )
		   at System.Dynamic.UpdateDelegates.UpdateAndExecute2[T0,T1,TRet](CallSite site, T0 arg0, T1 arg1)"
    TargetSite: {Void ThrowArgumentException(System.ExceptionResource)}
````

StefH:
Maybe also related to #634 ?

PR #631, opened 2022-09-16
Init field only on first execution

BBreiden:
The current getter throws if called more than once, as the dictionary has already been set up.

### DIFF
--- a/src/System.Linq.Dynamic.Core/DynamicClass.cs
+++ b/src/System.Linq.Dynamic.Core/DynamicClass.cs
@@ -18,24 +18,28 @@ namespace System.Linq.Dynamic.Core;
 /// </summary>
 public abstract class DynamicClass : DynamicObject
 {
-    private readonly Dictionary<string, object?> _propertiesDictionary = new();
+    private Dictionary<string, object?>? _propertiesDictionary = null;
 
     private Dictionary<string, object?> Properties
     {
         get
         {
-            foreach (PropertyInfo pi in GetType().GetProperties(BindingFlags.Public | BindingFlags.Instance))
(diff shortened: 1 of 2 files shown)


BBreiden:
This should solve issue #629

StefH (changes requested):
@BBreiden 

Can you please add a unit-test `DynamicClassTests.cs` which tests this fix?

BBreiden:
I've added a test. It fails without my change and works afterwards.

src/System.Linq.Dynamic.Core/DynamicClass.cs
```diff
@@ -18,24 +18,28 @@ namespace System.Linq.Dynamic.Core;
 /// </summary>
 public abstract class DynamicClass : DynamicObject
 {
-    private readonly Dictionary<string, object?> _propertiesDictionary = new();
+    private Dictionary<string, object?>? _propertiesDictionary = null;
```
StefH, line 21: why not create always an empty directory?

BBreiden: I guess this would also work, if the getter is also changed to check that the dictionary is empty before adding the entries. One just needs to avoid double entries somehow to fix the problem. Would you like me to change the PR?

BBreiden:
Suggested change wasn't approved.

StefH:
@BBreiden 
Can you recreate your PR?

(sorry ; I did not have enough time for this project...)

BBreiden:
I reopened the PR!
:-)

StefH (approved):

PR #658, opened 2023-01-03
Update ExpressionParser.cs

malegend:
Enable ignore case.


StefH:
Closing for now because it can have side-effects.

PR #687, opened 2023-03-07
Fix parsing a string literal with a dot

StefH:


jbhelm:
> 1.3.2-preview-01 is uploaded to https://www.myget.org/F/system-linq-dynamic-core/api/v3/index.json

Thank you!

Issue #702, opened 2023-05-02
GetAssemblyTypesWithDynamicLinqTypeAttribute fails to type load failures


philr:
For .NET Framework, [`AbstractDynamicLinqCustomTypeProvider.GetAssemblyTypesWithDynamicLinqTypeAttribute()` calls `Assembly.GetExportedTypes()`](https://github.com/zzzprojects/System.Linq.Dynamic.Core/blob/3e4262a8e4b4eed322722b1a5f011ae5138247a2/src/System.Linq.Dynamic.Core/CustomTypeProviders/AbstractDynamicLinqCustomTypeProvider.cs#L137) and handles a resulting `ReflectionTypeLoadException`.

[`Assembly.GetExportedTypes()`](https://learn.microsoft.com/dotnet/api/system.reflection.assembly.getexportedtypes) will actually throw `FileNotFoundException` if an assembly cannot be loaded.

`ReflectionTypeLoadException` is only thrown by [`Assembly.GetTypes()`](https://learn.microsoft.com/dotnet/api/system.reflection.assembly.gettypes).

As a result, if there's a failure loading a type caused by a missing assembly, none of the types marked with `DynamicLinqTypeAttribute` in the assembly will be returned.

StefH:
@philr Did you encounter this issue in your application, or did you check the source code?

Because the logic in the code is that if an assembly cannot be loaded, the `definedTypes` will be empty and no types will be returned for that assembly, but the next assembly will be analyzed.

StefH:
@philr 
If you have time, can you verify that PR?

philr:
@StefH I've added a review to the PR with a few comments.

PR #703, opened 2023-05-03
 Update logic for AbstractDynamicLinqCustomTypeProvider.GetAssemblyTypesWithDynamicLinqTypeAttribute

StefH:
https://github.com/zzzprojects/System.Linq.Dynamic.Core/issues/702

### DIFF
--- a/src/System.Linq.Dynamic.Core/CustomTypeProviders/AbstractDynamicLinqCustomTypeProvider.cs
+++ b/src/System.Linq.Dynamic.Core/CustomTypeProviders/AbstractDynamicLinqCustomTypeProvider.cs
@@ -83,14 +83,16 @@ public abstract class AbstractDynamicLinqCustomTypeProvider
     /// Gets the assembly types annotated with <see cref="DynamicLinqTypeAttribute"/> in an Exception friendly way.
     /// </summary>
     /// <param name="assemblies">The assemblies to process.</param>
-    /// <returns><see cref="IEnumerable{Type}" /></returns>
-    protected IEnumerable<Type> GetAssemblyTypesWithDynamicLinqTypeAttribute(IEnumerable<Assembly> assemblies)
+    /// <returns>Array of <see cref="Type" /></returns>
+    protected Type[] GetAssemblyTypesWithDynamicLinqTypeAttribute(IEnumerable<Assembly> assemblies)
     {
         Check.NotNull(assemblies);
 
+        var dynamicLinqTypes = new List<Type>();
+
         foreach (var assembly in assemblies)
         {
-            var definedTypes = Type.EmptyTypes;
+            Type[] definedTypes;
 
             try
             {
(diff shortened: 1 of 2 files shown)


src/System.Linq.Dynamic.Core/CustomTypeProviders/AbstractDynamicLinqCustomTypeProvider.cs
```diff
@@ -83,18 +83,20 @@ protected IEnumerable<Type> FindTypesMarkedWithDynamicLinqTypeAttribute(IEnumera
...
+
         foreach (var assembly in assemblies)
         {
-            var definedTypes = Type.EmptyTypes;
+            Type[] definedTypes;
 
             try
             {
-                definedTypes = assembly.ExportedTypes.ToArray();
+                definedTypes = assembly.GetExportedTypes();
```
philr, line 99: `GetExportedTypes()` still won't throw a `ReflectionTypeLoadException` according to the documentation. If there are individual types that can't be loaded, `definedTypes` will end up being set to `Type.EmptyTypes`.

src/System.Linq.Dynamic.Core/CustomTypeProviders/AbstractDynamicLinqCustomTypeProvider.cs
```diff
@@ -103,38 +105,49 @@ protected IEnumerable<Type> GetAssemblyTypesWithDynamicLinqTypeAttribute(IEnumer
...
+                    }
+                }
+                catch
+                {
+                    // Ignore
+                }
             }
         }
+
+        return dynamicLinqTypes.Distinct().ToArray();
```
philr, line 127: Is the `Distinct()` necessary? Each type should only be returned once from the Assembly.

src/System.Linq.Dynamic.Core/CustomTypeProviders/AbstractDynamicLinqCustomTypeProvider.cs
```diff
@@ -143,29 +156,26 @@ protected IEnumerable<Type> GetAssemblyTypesWithDynamicLinqTypeAttribute(IEnumer
...
                 }
             }
-
-            foreach (var definedType in filtered.Distinct().ToArray())
-            {
-                yield return definedType;
-            }
         }
+
+        return dynamicLinqTypes.Distinct().ToArray();
```
philr, line 178: Is the `Distinct()` necessary? Each type should only be returned once from the Assembly.

philr (commented):
For .NET Framework, this is changing from calling `assembly.GetExportedTypes()` to `assembly.GetTypes()`. That allows the `ReflectionTypeLoadException` to be handled, but changes the behaviour to returning all the types in the assembly instead of just the public types. I don't know if you were intending change that? If not, the result can be filtered down to just the public types.

.NET Standard changes from `assembly.ExportedTypes` to `assembly.GetExportedTypes()`. `GetExportedTypes()` won't ever throw `ReflectionTypeLoadException` according to the documentation. This should be changed to `assembly.GetTypes()` like the .NET Framework version of the code (and potentially filtered to just public types)?

PR #723, opened 2023-07-11
Handle UseParameterizedNamesInDynamicQuery when parsing dynamic OfType function

### LINKED ISSUE #722: UseParameterizedNamesInDynamicQuery set to true breaks `OfType` method call
## 1. Description

We are working with the lib to generate linq queries from strings and we recently found the setting `UseParameterizedNamesInDynamicQuery`, which we see as a nice speed optimization. We updated our parameters to set that to true, and now we have a problem with the `.OfType("type")` method - `System.Linq.Dynamic.Core.Exceptions.ParseException: The 'OfType' function requires the argument to be not null and of type ConstantExpression.`

Our usage looks like this (abbreviated a whole lot for simplicity):

```csharp
var queryable = dataContext.GetEntity<EntityType>();
queryable.Where($"Field.OfType(\"{typeof(MyType)}\")", args);
```

## 2. Exception
If you are seeing an exception, include the full exception details (message and stack trace).

```
Exception message: System.Linq.Dynamic.Core.Exceptions.ParseException: The 'OfType' function requires the argument to be not null and of type ConstantExpression.

Stack trace: ExpressionParser.ResolveTypeFromArgumentExpression(String functionName, Expression argumentExpression, Nullable`1 arguments)
ExpressionParser.ParseEnumerable(Expression instance, Type elementType, String methodName, Int32 errorPos, Type type)
ExpressionParser.ParseMemberAccess(Type type, Expression expression)
ExpressionParser.ParsePrimary()
ExpressionParser.ParseUnary()
ExpressionParser.ParseMultiplicative()
ExpressionParser.ParseAdditive()
ExpressionParser.ParseShiftOperator()
ExpressionParser.ParseComparisonOperator()
ExpressionParser.ParseLogicalAndOrOperator()
<21 more frames...>
ExpressionParser.Parse(Type resultType, Boolean createParameterCtor)
DynamicExpressionParser.ParseLambda(Type delegateType, ParsingConfig parsingConfig, Boolean createParameterCtor, ParameterExpression[] parameters, Type resultType, String expression, Object[] values)
DynamicExpressionParser.ParseLambda(ParsingConfig parsingConfig, Boolean createParameterCtor, ParameterExpression[] parameters, Type resultType, String expression, Object[] values)
DynamicExpressionParser.ParseLambda(ParsingConfig parsingConfig, Boolean createParameterCtor, Type itType, Type resultType, String expression, Object[] values)
DynamicQueryableExtensions.Where(IQueryable source, ParsingConfig config, String predicate, Object[] args)
DynamicQueryableExtensions.Where[TSource](IQueryable`1 source, ParsingConfig config, String predicate, Object[] args)
```

## 3. Fiddle or Project

I am not quite able to setup a reproduction in Fiddle right now, nor provide a (private) project - this is really deep into convoluted code - I'll look further into this if it is required to help you figure out the problem.

Our usage looks like this (abbreviated a whole lot for simplicity):

```csharp
var parsingConfig = new ParsingConfig { UseParameterizedNamesInDynamicQuery = true };
var queryable = dataContext.GetEntity<MyType>();
// build up a query here which provides arguments, ignored for this base example
queryable.Where(parsingConfig, $"Field.OfType(\"{typeof(MySubType)}\")", args);
```

## 4. Any further technical details

I dug around in the code and from what I could understand setting `UseParameterizedNamesInDynamicQuery` to `true` makes it so that the `Expression` that ends up reaching [`ExpressionParser.ResolveTypeFromArgumentExpression`](https://github.com/zzzprojects/System.Linq.Dynamic.Core/blob/7b94ef5b9536c9cbca5644dda14a74c9528ec178/src/System.Linq.Dynamic.Core/Parser/ExpressionParser.cs#L2006) is a `MemberExpression` instead of the expected `ConstantExpression`, making the [`switch (argumentExpression)`](https://github.com/zzzprojects/System.Linq.Dynamic.Core/blob/7b94ef5b9536c9cbca5644dda14a74c9528ec178/src/System.Linq.Dynamic.Core/Parser/ExpressionParser.cs#L2009) fall into the default case which throws the above exception.

To confirm my hypothesis I did a dirty check right before the `switch`:

```csharp
private Type ResolveTypeFromArgumentExpression(string functionName, Expression argumentExpression, int? arguments = null)
{
// ...
var unwrappedExpression = argumentExpression as ConstantExpression;
if (unwrappedExpression == null)
{
    _expressionHelper.TryUnwrapAsConstantExpression(argumentExpression, out unwrappedExpression);
}

switch (unwrappedExpression)
// ...
```

and that fixed the problem we were hitting with the `OfType` method.

yonguelink:
Fixes #722

### DIFF
--- a/src/System.Linq.Dynamic.Core/Parser/ExpressionParser.cs
+++ b/src/System.Linq.Dynamic.Core/Parser/ExpressionParser.cs
@@ -2013,16 +2013,20 @@ public class ExpressionParser
                 return ResolveTypeFromExpressionValue(functionName, constantExpression, argument);
 
             case MemberExpression memberExpression:
-                _expressionHelper.TryUnwrapAsConstantExpression(memberExpression, out var unwrappedConstantExpression);
-                return ResolveTypeFromExpressionValue(functionName, unwrappedConstantExpression, argument);
-            default:
-                throw ParseError(_textParser.CurrentToken.Pos, Res.FunctionRequiresNotNullArgOfType, functionName, argument, "ConstantExpression");
+                if (_expressionHelper.TryUnwrapAsConstantExpression(memberExpression, out var unwrappedConstantExpression))
+                {
+                    return ResolveTypeFromExpressionValue(functionName, unwrappedConstantExpression, argument);
+                }
+
+                break;
         }
+
+        throw ParseError(_textParser.CurrentToken.Pos, Res.FunctionRequiresNotNullArgOfType, functionName, argument, "ConstantExpression");
     }
 
-    private Type ResolveTypeFromExpressionValue(string functionName, ConstantExpression? constantExpression, string argument)
+    private Type ResolveTypeFromExpressionValue(string functionName, ConstantExpression constantExpression, string argument)
     {
-        switch (constantExpression?.Value)
+        switch (constantExpression.Value)
         {
             case string typeName:
                 return ResolveTypeStringFromArgument(typeName);
--- a/src/System.Linq.Dynamic.Core/Parser/ExpressionParser.cs
+++ b/src/System.Linq.Dynamic.Core/Parser/ExpressionParser.cs
@@ -2006,23 +2006,32 @@ public class ExpressionParser
     private Type ResolveTypeFromArgumentExpression(string functionName, Expression argumentExpression, int? arguments = null)
     {
         string argument = arguments == null ? string.Empty : arguments == 1 ? "first " : "second ";
+
         switch (argumentExpression)
         {
             case ConstantExpression constantExpression:
-                switch (constantExpression.Value)
-                {
-                    case string typeName:
-                        return ResolveTypeStringFromArgument(typeName);
+                return ResolveTypeFromExpressionValue(functionName, constantExpression, argument);
 
-                    case Type type:
-                        return type;
-
-                    default:
-                        throw ParseError(_textParser.CurrentToken.Pos, Res.FunctionRequiresNotNullArgOfType, functionName, argument, "string or System.Type");
-                }
-
+            case MemberExpression memberExpression:
+                _expressionHelper.TryUnwrapAsConstantExpression(memberExpression, out var unwrappedConstantExpression);
+                return ResolveTypeFromExpressionValue(functionName, unwrappedConstantExpression, argument);
             default:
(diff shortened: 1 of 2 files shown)


src/System.Linq.Dynamic.Core/Parser/ExpressionParser.cs
```diff
@@ -2006,23 +2006,32 @@ private Expression ParseEnumerable(Expression instance, Type elementType, string
...
         {
             case ConstantExpression constantExpression:
-                switch (constantExpression.Value)
-                {
-                    case string typeName:
-                        return ResolveTypeStringFromArgument(typeName);
+                return ResolveTypeFromExpressionValue(functionName, constantExpression, argument);
+
+            case MemberExpression memberExpression:
+                _expressionHelper.TryUnwrapAsConstantExpression(memberExpression, out var unwrappedConstantExpression);
```
StefH, line 2016: The TryUnwrapAsConstantExpression returns a bool indicating it's valid.

This code does not check it ?

StefH (changes requested):

src/System.Linq.Dynamic.Core/Parser/ExpressionParser.cs
yonguelink: I started working with that if, but I would've either needed to duplicate the `throw ParseError` from below OR remove the `default` case from the `switch` & throw there.

I decided to ignore the boolean to make things straightforward, letting the `ResolveTypeFromExpressionValue` handle the potential null `unwrappedConstantExpression` out var (line 2025)

If you prefer the approach with the bool check, I can do that, but which one would you prefer?

- Dropping the `default` case from the `switch` to avoid duplicating the `throw` statement
- Duplicating the `throw` statement in the `MemberExpression` case of the switch

yonguelink: I went with the first option to drop the `default` case while checking for the `TryUnwrapAsConstantExpression` return value instead of propagating potential null/default values.

That makes more sense to me.

StefH (approved):

yonguelink:
@StefH is there something else to do to get this completed?

StefH:
@yonguelink 
It's good. Thank you.

I'll merge this to master.

Issue #764, opened 2024-01-19
Possible Memory Leak with ConstantExpressionHelper


TWhidden:
**Description**
We are currently investigating a growing leak in our net8 app.  Doing a `dotnet-dump`, and using JetBrains `dotMemory`, we found over 761k objects in the `ConcurrentDictionary<object,Expression>`, and 695k `ConcurrentDictionary<Expression, string>` both found in the `ConstantExpressionHelper`. We believe this is what is causing our leak.  Analysis of all the instances show this class overwhelming on top of any other type instance.

**Usage:**
Our application makes heavy use of the dynamic library - where expression trees are created from input string data provided by outside calls.  This has worked well for many years, going back to the early introduction of the dynamic linq library in the framework days.

Inputs change very regularly, based on the applications requirements. Constants are introduced that may only be used once in the application lifetime.

**Possible Solution:**
After cloning the project, was able to see the `ConcurrentDictionary` in `ConstantExpressionHelper` grow without any way to remove collected constant expressions.  Proposal would be to use a `sliding cache` with a reasonable short lifetime (minutes? hours?), instead of an ever-lasting private static `ConcurrentDictionary`, allowing the system to remove constants that are no longer being called, which will then allow the GC to remove the tress of referenced objects. 

**Instance Details**
Process running under Ubuntu Linux Container, net8.0 framework, approximately 9 days of uptime.

Trends monitoring:
![image](https://github.com/zzzprojects/System.Linq.Dynamic.Core/assets/9921122/260ffda9-f070-43ae-82da-58b2b0825b46)

Screen Shots of dotMemory outputs, arranged by size descending:
![image](https://github.com/zzzprojects/System.Linq.Dynamic.Core/assets/9921122/b7aed8ff-c4dd-4ff2-983e-2e6847df0133)

Event showing 18 megs of Int32 constants
![image](https://github.com/zzzprojects/System.Linq.Dynamic.Core/assets/9921122/48a7e336-d97e-41cf-aa28-d6563f0fbdc7)

The use of the `ConstantExpressionHelper` makes total sense, so I can appreciate why it's there. The proposal is just to have a way to let it clear itself out, especially for highly dynamic, long running services.  Self-managing is desired, no reason for the developer to write code responsible to clear these things out. 

[...]

PR #765, opened 2024-01-19
#764 - Introduce Sliding Cache to Constant Expression Helper

TWhidden:
Introduce a Sliding Cache to ConstantExpressionHelper.  
- Use of DateTime (UTC) instead of Stopwatch for expire period. lightweight
- Should still work with older frameworks - Still using same backing store, just with a private Tuple that holds the last accessed time
- Introduced new internal type `ThreadSafeSlidingCache` which has a constructor option of TTL for the object expire period, as well as an override for the cleanup frequency - default of 10 seconds atm - figured that would not add a huge strain, but can be adjusted as desired. did not know where we would put this class in the folder structure, feel free to refactor. It uses a "Lazy" Expire, so there is no timer, but objects expire when the cache is accessed. 
- TTL default for objects built in Release is sliding cache value of 10 minutes. This is just an arbitrary number that seems reasonable. 
- Test Case created - Passes on my end - feel free to refactor the name of the test


src/System.Linq.Dynamic.Core/Parser/ThreadSafeSlidingCache.cs
```diff
@@ -0,0 +1,80 @@
+﻿using System.Collections.Concurrent;
+
+namespace System.Linq.Dynamic.Core.Parser
+{
+    internal class ThreadSafeSlidingCache<T1, T2> where T1 : notnull where T2 : notnull
```
StefH, line 5: Rename to TKey

TWhidden: Will do the same for `TValue`

StefH (changes requested):
Thanks for this PR.

Can you look at the review comments please?

src/System.Linq.Dynamic.Core/Parser/ThreadSafeSlidingCache.cs
```diff
@@ -0,0 +1,80 @@
...
+        private readonly ConcurrentDictionary<T1, (T2 Value, DateTime ExpirationTime)> _cache;
+        private readonly TimeSpan _timeToLive;
+        private readonly TimeSpan _cleanupFrequency;
+        private DateTime _lastCleanupTime = DateTime.MinValue;
+
+        public ThreadSafeSlidingCache(TimeSpan timeToLive, TimeSpan? cleanupFrequency = null)
+        {
+            _cache = new ConcurrentDictionary<T1, (T2, DateTime)>();
+            _timeToLive = timeToLive;
+            _cleanupFrequency = cleanupFrequency ?? TimeSpan.FromSeconds(10);
```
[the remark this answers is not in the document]
TWhidden: Resharper gives a minor concern, `Static Member in Generic Type` - do you permit these kinds of annotations in the source?
`// ReSharper disable once StaticMemberInGenericType`

src/System.Linq.Dynamic.Core/Parser/ThreadSafeSlidingCache.cs
```diff
@@ -0,0 +1,80 @@
...
+
+        public void AddOrUpdate(T1 key, T2 value)
+        {
+            if (key == null) throw new ArgumentNullException(nameof(key));
+            if (value == null) throw new ArgumentNullException(nameof(value));
+
+            var expirationTime = DateTime.UtcNow.Add(_timeToLive);
+            _cache[key] = (value, expirationTime);
+
+            CleanupIfNeeded();
```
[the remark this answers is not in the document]
TWhidden: Excellent idea. Great source reference. Will prevent the block.

src/System.Linq.Dynamic.Core/Parser/ThreadSafeSlidingCache.cs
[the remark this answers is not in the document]
StefH: One way to solve this could be to have 2 classes: see example below....

``` c#
namespace WireMock.Util;

internal static class SingletonLock
{
    public static readonly object Lock = new();
}

internal static class SingletonFactory<T> where T : class, new()
{
    private static T? _instance;

    public static T GetInstance()
    {
        if (_instance == null)
        {
            lock (SingletonLock.Lock)
            {
                _instance ??= new T();
            }
        }

        return _instance;
    }
}
```

src/System.Linq.Dynamic.Core/Parser/ConstantExpressionHelper.cs
```diff
@@ -1,12 +1,18 @@
...
+﻿using System.Linq.Expressions;
 
 namespace System.Linq.Dynamic.Core.Parser
 {
     internal static class ConstantExpressionHelper
     {
-        private static readonly ConcurrentDictionary<object, Expression> Expressions = new();
-        private static readonly ConcurrentDictionary<Expression, string> Literals = new();
+#if DEBUG
+        private static readonly TimeSpan TimeToLivePeriod = TimeSpan.FromSeconds(10);
```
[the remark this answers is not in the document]
TWhidden: I recommend keeping `ConstantExpressionHelper` as a static, non-configurable class. Its widespread use throughout the library suggests that refactoring it to manage a static instance elsewhere would involve significant changes. The current implementation seems to be the most straightforward and minimizes potential misuse.

Additionally, the simplicity of the Sliding Cache within `ConstantExpressionHelper` should be maintained. Overcomplicating it with configurations could lead to unnecessary complexity. The library itself handles the cleanup, which should remain transparent to the users. Just my .02c.

src/System.Linq.Dynamic.Core/Parser/ThreadSafeSlidingCache.cs
```diff
@@ -0,0 +1,80 @@
...
+namespace System.Linq.Dynamic.Core.Parser
+{
+    internal class ThreadSafeSlidingCache<T1, T2> where T1 : notnull where T2 : notnull
+    {
+        private readonly ConcurrentDictionary<T1, (T2 Value, DateTime ExpirationTime)> _cache;
+        private readonly TimeSpan _timeToLive;
+        private readonly TimeSpan _cleanupFrequency;
+        private DateTime _lastCleanupTime = DateTime.MinValue;
+
+        public ThreadSafeSlidingCache(TimeSpan timeToLive, TimeSpan? cleanupFrequency = null)
```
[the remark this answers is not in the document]
TWhidden: For the ConstantExpressionHelper, I believe that specifying the maximum number is not a critical concern for developers, given the unpredictability and variability of different projects. However, considering its potential utility as an aid to other parts of the library in the future, adding an optional configuration for this could be beneficial. Therefore, I have included it as an optional parameter in the constructor.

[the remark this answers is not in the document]
TWhidden: Great idea on the DateTime Mock. Implemented. I added `IDateTimeUtils` under Util, hopefully that is ok. (internal)

**Unit Testing:** New test cases for `ThreadSafeSlidingCache` have been added under the 'Util' folder in our test suite. These tests cover:
   - Adding, updating, removal operations.
   - `TryGetValue` functionality.
   - Expiry mechanism testing, including both manual expiry and automatic expiry scenarios (e.g., attempting to access a non-existent key and observing the auto-expiration of an existing key)."

src/System.Linq.Dynamic.Core/Parser/ThreadSafeSlidingCache.cs
```diff
@@ -0,0 +1,83 @@
...
+            _cache = new ConcurrentDictionary<TKey, (TValue, DateTime)>();
+            _timeToLive = timeToLive;
+            _cleanupFrequency = cleanupFrequency ?? TimeSpan.FromSeconds(10);
+        }
+
+        public TimeSpan TimeToLive => _timeToLive;
+
+        public void AddOrUpdate(TKey key, TValue value)
+        {
+            if (key == null) throw new ArgumentNullException(nameof(key));
```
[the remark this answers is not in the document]
TWhidden: Couldn't find usage of `Guard.NotNull` in the library, but did find `Check.NotNull(` - I assume that is what you wanted to use? Added a unit test for this also and throws correctly. `System.ArgumentNullException : Value cannot be null. (Parameter 'key')`

src/System.Linq.Dynamic.Core/Parser/ThreadSafeSlidingCache.cs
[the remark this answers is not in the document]
TWhidden: Excellent idea - ugg, Sunday morning, I need a redbull!  Updated.

src/System.Linq.Dynamic.Core/Parser/ThreadSafeSlidingCache.cs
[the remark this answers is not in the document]
StefH: > Couldn't find usage of `Guard.NotNull` in the library, but did find `Check.NotNull(` - I assume that is what you wanted to use? Added a unit test for this also and throws correctly. `System.ArgumentNullException : Value cannot be null. (Parameter 'key')`

**Check.NotNull** is the correct one indeed

TWhidden:
@StefH  - Ok, implemented the feedback above, with the exception of a configuration process as I think that really isn't needed for this (not sure who would be trying to tune this, this is just the library keeping itself clean so the GC can do its job).  

 - Added a Unit Test for the `ThreadSafeSlidingCache` and all its functions.
 - Removed the `ConstantExpressionHelper` since the unit tests for the `ThreadSafeSlidingCache` should be trusted. 
 - Dropped pre-processor directive I was using for tests, since implementing the Mock DateTime provider.

Thanks for the idea of the Mock DateTime - I can use that a lot on my own project tests!

test/System.Linq.Dynamic.Core.Tests/Util/ThreadSafeSlidingCacheTests.cs
```diff
@@ -0,0 +1,152 @@
...
+
+            // Since the cache cleanup is triggered by a Task and not on the same thread, 
+            // give it a moment for the cleanup to happen
+            await Task.Delay(10);
+
+            // Ensure one item is in the cache
+            Assert.True(cache.Count == 1, $"Expected 1 items in the cache, had {cache.Count}");
+        }
+
+        private class MockDateTimeProvider : IDateTimeUtils
```
StefH, line 146: With this logic, you are still using a real DateTime.

The correct way to do is, is to Mock it:
``` c#
private static readonly DateTime UtcNow = new(2023, 1, 14, 15, 16, 17);

var dateTimeUtilsMock = new Mock<IDateTimeUtils>();
dateTimeUtilsMock.SetupGet(d => d.UtcNow).Returns(UtcNow);
```

See this test class:
https://github.com/WireMock-Net/WireMock.Net/blob/13a3dfa9feef7322064145a82bfe4ede50ef8974/test/WireMock.Net.Tests/MappingBuilderTests.cs#L45

test/System.Linq.Dynamic.Core.Tests/Util/ThreadSafeSlidingCacheTests.cs
```diff
@@ -0,0 +1,152 @@
...
+                dateTimeProvider: mockDateTime);
+
+            // Act
+            cache.AddOrUpdate(1, "one");
+
+            // Ensure one item is in the cache
+            Assert.True(cache.Count == 1, $"Expected 1 items in the cache, only had {cache.Count}");
+
+            // move the time forward
+            mockDateTime.UtcNow = mockDateTime.UtcNow.AddMinutes(11);
```
StefH, line 123: Use **SetupSequence**


Example
``` c#
    public class YourTestClass
    {
        public void YourTestMethod()
        {
            // Create a mock of IDateTimeUtils
            var mockDateTimeUtils = new Mock<IDateTimeUtils>();

            // Set up a sequence for UtcNow to return different values
            mockDateTimeUtils.SetupSequence(x => x.UtcNow)
                             .Returns(new DateTime(2022, 1, 1))
                             .Returns(new DateTime(2022, 1, 2))
                             .Returns(new DateTime(2022, 1, 3));

            // Example usage
            Console.WriteLine(mockDateTimeUtils.Object.UtcNow); // Outputs 2022-01-01
            Console.WriteLine(mockDateTimeUtils.Object.UtcNow); // Outputs 2022-01-02
            Console.WriteLine(mockDateTimeUtils.Object.UtcNow); // Outputs 2022-01-03
        }
    }

```

test/System.Linq.Dynamic.Core.Tests/Util/ThreadSafeSlidingCacheTests.cs
```diff
@@ -0,0 +1,152 @@
...
+
+            // Add
+            cache.AddOrUpdate(1, "one");
+            cache.AddOrUpdate(2, "two");
+            cache.AddOrUpdate(3, "three");
+
+            // Replace
+            cache.AddOrUpdate(1, "oneone");
+
+            Assert.True(cache.Count == 3, $"Expected 3 items in the cache, only had {cache.Count}");
```
StefH, line 28: Do not use `Assert.True` --> that is only used to check for a true boolean value.

If you want to check a normal value, use:
``` c#
Assert.Equal(3, cache.Count, $"Expected 3 items in the cache, only had {cache.Count}");
```

Or my personal preference : FluentAssertions:
``` c#
cache.Count.Should().Be(3, $"Expected 3 items in the cache, only had {cache.Count}");
```

test/System.Linq.Dynamic.Core.Tests/Util/ThreadSafeSlidingCacheTests.cs
```diff
@@ -0,0 +1,152 @@
...
+            // Arrange
+            var cache = new ThreadSafeSlidingCache<int, string>(TimeSpan.FromMinutes(10),
+                dateTimeProvider: mockDateTime);
+
+            // Act
+            cache.AddOrUpdate(1, "one");
+            mockDateTime.UtcNow = mockDateTime.UtcNow.AddMinutes(11);
+            if (cache.TryGetValue(1, out var value))
+            {
+                Assert.True(false, $"Expected to not find the value, but found {value}");
```
StefH, line 56: In this case, use `Asser.Fail`

src/System.Linq.Dynamic.Core/Util/ThreadSafeSlidingCache.cs
```diff
@@ -0,0 +1,135 @@
...
+
+                Task.Run(_deleteExpiredCachedItemsDelegate);
+            }
+        }
+
+        /// <summary>
+        ///     Cleanup the Cache
+        /// </summary>
+        /// <returns></returns>
+        private Task Cleanup()
```
StefH, line 122: Rename to CleanupAsync

src/System.Linq.Dynamic.Core/Util/ThreadSafeSlidingCache.cs
```diff
@@ -0,0 +1,135 @@
...
+               )
+            {
+                // Set here, so we don't have re-entry due to large collection enumeration.
+                _lastCleanupTime = _dateTimeProvider.UtcNow;
+
+                Task.Run(_deleteExpiredCachedItemsDelegate);
+            }
+        }
+
+        /// <summary>
```
StefH, line 118: private methods do not need xml-comment, expect when very special or complex

src/System.Linq.Dynamic.Core/Util/ThreadSafeSlidingCache.cs
```diff
@@ -0,0 +1,135 @@
...
+        // Default cleanup frequency
+        public static readonly TimeSpan DefaultCleanupFrequency = TimeSpan.FromMinutes(10);
+    }
+
+    internal class ThreadSafeSlidingCache<TKey, TValue> where TKey : notnull where TValue : notnull
+    {
+        private readonly ConcurrentDictionary<TKey, (TValue Value, DateTime ExpirationTime)> _cache;
+        private readonly TimeSpan _cleanupFrequency;
+        private readonly IDateTimeUtils _dateTimeProvider;
+        private readonly Func<Task> _deleteExpiredCachedItemsDelegate;
```
StefH, line 18: In this case, I think that this can be an `Action` (so the method is just a simple void method).

src/System.Linq.Dynamic.Core/Util/ThreadSafeSlidingCache.cs
```diff
@@ -0,0 +1,135 @@
...
+            Check.NotNull(key);
+
+            var removed = _cache.TryRemove(key, out _);
+            CleanupIfNeeded();
+            return removed;
+        }
+
+        /// <summary>
+        ///     Check if cache needs to be cleaned up.
+        ///     If it does, span the cleanup as a Task to prevent from blocking
```
StefH, line 101: `If it does, start the cleanup as a Task to prevent from blocking`

src/System.Linq.Dynamic.Core/Util/ThreadSafeSlidingCache.cs
```diff
@@ -0,0 +1,135 @@
...
+{
+    internal static class ThreadSafeSlidingCacheConstants
+    {
+        // Default cleanup frequency
+        public static readonly TimeSpan DefaultCleanupFrequency = TimeSpan.FromMinutes(10);
+    }
+
+    internal class ThreadSafeSlidingCache<TKey, TValue> where TKey : notnull where TValue : notnull
+    {
+        private readonly ConcurrentDictionary<TKey, (TValue Value, DateTime ExpirationTime)> _cache;
```
StefH, line 15: "Named": Tuple is not supported in older frameworks like 4.5.x
image

So use `Tuple<TValue, DateTime>` or create a simple custom struct for this.

(Maybe create a sub-folder in "Utils" named "Cache", where you can place this file and also move ThreadSafeSlidingCacheConstants to an own class in that new folder.

src/System.Linq.Dynamic.Core/Util/ThreadSafeSlidingCache.cs
```diff
@@ -0,0 +1,135 @@
...
+            Check.NotNull(key);
+            Check.NotNull(value);
+
+            var expirationTime = _dateTimeProvider.UtcNow.Add(TimeToLive);
+            _cache[key] = (value, expirationTime);
+
+            CleanupIfNeeded();
+        }
+
+        public bool TryGetValue(TKey key, out TValue value)
```
StefH, line 67: Officially this Try should look like this example:

- `out TValue? value`
- `[NotNullWhen(true)] `

```` c#
public bool TryUnwrapAsValue<TValue>(Expression? expression, [NotNullWhen(true)] out TValue? value)
    {
        if (_parsingConfig.UseParameterizedNamesInDynamicQuery && _constantExpressionWrapper.TryUnwrapAsValue(expression as MemberExpression, out value))
        {
            return true;
        }

        value = default;   // <--------- just default, not default!
        return false;
    }
````

File: src/System.Linq.Dynamic.Core/Util/ThreadSafeSlidingCache.cs [hunk 17]
```diff
@@ -0,0 +1,135 @@
...
+            if (_dateTimeProvider.UtcNow - _lastCleanupTime > _cleanupFrequency
+                && (_minCacheItemsBeforeCleanup == null ||
+                    _cache.Count >=
+                    _minCacheItemsBeforeCleanup) // Only cleanup if we have a minimum number of items in the cache.
+               )
+            {
+                // Set here, so we don't have re-entry due to large collection enumeration.
+                _lastCleanupTime = _dateTimeProvider.UtcNow;
+
+                Task.Run(_deleteExpiredCachedItemsDelegate);
```
StefH, line 114: `Task.Run` does not exists for .NET 3.5

Make a new static class named `TaskUtils` and add a method `Run` which looks the same as Task.Run (with an `Action action`).

Add a preprocessor directive for .NET35 to use code like:
``` c#
ThreadPool.QueueUserWorkItem(_ => _deleteExpiredCachedItemsDelegate.Invoke());
```` 

And Else just use Task.Run

**Review by StefH — CHANGES_REQUESTED** [review]

File: test/System.Linq.Dynamic.Core.Tests/Util/ThreadSafeSlidingCacheTests.cs [hunk 18]
```diff
@@ -0,0 +1,152 @@
+﻿using System.Linq.Dynamic.Core.Util;
+using System.Linq.Expressions;
+using System.Threading.Tasks;
+using Xunit;
+
+namespace System.Linq.Dynamic.Core.Tests.Util
+{
```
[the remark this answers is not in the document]
TWhidden: I **love** file-scoped namespaces, but the csproj `LangVersion` is not defined (or set to `Latest`). When I use Resharper to cleanup, it re-adds the scope brackets.  I manually set it for now, but may I suggest the `LangVersion` be defined in the `csproj` so things like the RS Code Cleanup will auto-refactor that?  

https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/configure-language-version 

Not sure why RS doesn't honor C#12 by default for 8 projects, maybe because it has several Target Frameworks? not sure.  Let me know if I can add that in.

test/System.Linq.Dynamic.Core.Tests/Util/ConstantExpressionHelperTests.cs
```diff
@@ -0,0 +1,48 @@
+﻿using System.Linq.Dynamic.Core.Parser;
```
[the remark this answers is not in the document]
TWhidden: I deleted it - Don't think it's needed any-longer with the new Test for the Sliding Cache we added. Ill add that to the next check-in.

test/System.Linq.Dynamic.Core.Tests/Util/ThreadSafeSlidingCacheTests.cs
TWhidden: Implementing - I don't have any experience with that library - what is the proper way I should advance the time in my tests?  

I did this - seems to work, but not sure if there is a cleaner way to do it.

```
var r = dateTimeUtilsMock.Object.UtcNow.AddMinutes(11);
dateTimeUtilsMock.SetupGet(d => d.UtcNow).Returns(r);
```

test/System.Linq.Dynamic.Core.Tests/Util/ThreadSafeSlidingCacheTests.cs
TWhidden: Oh sorry - missed this - disregard my last question. This makes more sense.

test/System.Linq.Dynamic.Core.Tests/Util/ThreadSafeSlidingCacheTests.cs
TWhidden: Having trouble with `Assert.Equal` when including a description. 
the `FluentAssertions` is quite slick though, going with that.

test/System.Linq.Dynamic.Core.Tests/Util/ThreadSafeSlidingCacheTests.cs
TWhidden: Lol, that's a Co-pilot fail!  Fixed! How did I miss that!

src/System.Linq.Dynamic.Core/Util/ThreadSafeSlidingCache.cs
TWhidden: In this case, will not add the `Async` to the end, since it's not async any longer.  Agree, this is probably better.

src/System.Linq.Dynamic.Core/Util/ThreadSafeSlidingCache.cs
TWhidden: Probably would have picked all this up if I installed all the other frameworks on my dev system. Just hate having to install such legacy targets! I'll try and get that done tonight. I am not a huge fan of named tuples just yet, but it's very convenient for private items like this - but easy to swap, and kind of cleaner without.

src/System.Linq.Dynamic.Core/Util/ThreadSafeSlidingCache.cs
TWhidden: Agree - a nice pattern especially when using nullable!

test/System.Linq.Dynamic.Core.Tests/Util/ThreadSafeSlidingCacheTests.cs
TWhidden: So, going back to this - having trouble with xUnit - does not appear they have `Assert.Fail(..)` - I think Co-pilot suggested that as an alternative.  Some StackOverflow articles show I should just throw an exception instead, which will fail it. I am not sure how I feel about that.  How would you want me to trigger a failure?

test/System.Linq.Dynamic.Core.Tests/Util/ThreadSafeSlidingCacheTests.cs
TWhidden: I have a question regarding the cache class provider's frequent internal usage. To simplify the process and avoid the complexity of tracking the sequence of internal accesses, would it be more efficient to directly set the new time as needed? I'm concerned that the test class might become overly complicated if the cache is accessed multiple times internally before advancing the time. There are numerous internal comparisons, and detailing these in the code seems cumbersome. For my purposes, would it be feasible to implement a direct time-setting approach?

```
var r = dateTimeUtilsMock.Object.UtcNow.AddMinutes(11);
dateTimeUtilsMock.SetupGet(d => d.UtcNow).Returns(r);
```

TWhidden:
**Feedback on the Latest Check-In:**

1. **Singleton Implementation in `ConstantExpressionHelper`:** I've implemented an instance of `ConstantExpressionHelper` using a singleton pattern to prevent type duplication across instances. I'd appreciate any feedback on this pattern.

2. **Testing - Setup Sequence:** Seeking feedback on the setup sequence for tests. Setting up every datetime feels cumbersome, as the provider is extensively used within the Cache. Is it necessary to detail every access in the SetupSequence, or is setting the next time sufficient for these tests?

3. **Issues with .NET Framework and Project Configurations:**
   - Encountered issues with the .NET 4.0 targeting pack and .NET 4.5.2 compilation, unrelated to my changes. These issues stemmed from other types within the class.
   - I had to unload projects for .NET 3.5 (concerns about `TaskUtil`) and disable Android and UWP tests due to signing and certificate issues. My environment might require additional configuration for these tests.

4. **New ParserConfig Options:** Added three new `ParserConfig` options to reflect the different possibilities within the Cache of `ConstantExpressionHelper`, as suggested.

src/System.Linq.Dynamic.Core/Parser/ConstantExpressionHelper.cs
```diff
@@ -1,29 +1,60 @@
...
+
+    // Static shared instance to prevent duplications of the same objects
+    private static ThreadSafeSlidingCache<object, Expression>? _expressions;
+    private static ThreadSafeSlidingCache<Expression, string>? _literals;
+
+    public ConstantExpressionHelper(ParsingConfig config)
     {
-        private static readonly ConcurrentDictionary<object, Expression> Expressions = new();
-        private static readonly ConcurrentDictionary<Expression, string> Literals = new();
+        _config = config;
```
StefH, line 16: `_config = Check.NotNull(config);`

src/System.Linq.Dynamic.Core/Parser/ConstantExpressionHelper.cs
```diff
@@ -1,29 +1,60 @@
...
+    private ThreadSafeSlidingCache<object, Expression> GetExpression()
+    {
+        _expressions ??= new ThreadSafeSlidingCache<object, Expression>(
+            _config.ConstantExpressionSlidingCacheTimeToLive,
+            _config.ConstantExpressionSlidingCacheCleanupFrequency,
+            _config.ConstantExpressionSlidingCacheMinItemsTrigger
+        );
+        return _expressions;
+    }
+
```
StefH, line 39: newline

src/System.Linq.Dynamic.Core/Parser/ConstantExpressionHelper.cs
```diff
@@ -1,29 +1,60 @@
...
-namespace System.Linq.Dynamic.Core.Parser
+namespace System.Linq.Dynamic.Core.Parser;
+
+internal class ConstantExpressionHelper
 {
-    internal static class ConstantExpressionHelper
+    private readonly ParsingConfig _config;
+
+    // Static shared instance to prevent duplications of the same objects
+    private static ThreadSafeSlidingCache<object, Expression>? _expressions;
```
StefH, line 11: In this case I think it's better to use a SingletonFactory pattern which takes care that the ConstantExpressionHelper is only created once.

Something like this:
``` c#
internal static class ConstantExpressionHelperFactory
{
    private static readonly object Lock = new();
    private static ConstantExpressionHelper? _instance;

    public static ConstantExpressionHelper GetInstance(ParsingConfig config)
    {
        if (_instance == null)
        {
            lock (Lock)
            {
                _instance ??= new ConstantExpressionHelper(config);
            }
        }

        return _instance;
    }
}
```

Based on this code:
https://github.com/WireMock-Net/WireMock.Net/blob/stef-protobuf/src/WireMock.Net/Util/SingletonFactory.cs

src/System.Linq.Dynamic.Core/Util/TaskUtils.cs
```diff
@@ -0,0 +1,16 @@
+﻿namespace System.Linq.Dynamic.Core.Util;
+
+internal static class TaskUtils
+{
+    public static void Run(Action action)
+    {
+#if NET35 || NET40
+        System.Threading.ThreadPool.QueueUserWorkItem(_ => {
+            action?.Invoke();
```
StefH, line 9: 1. This can be a one-liner:
```
 System.Threading.ThreadPool.QueueUserWorkItem(_ => action.Invoke());
```

2. No need to use `?` if you just add a Check.NotNull on line 6.

src/System.Linq.Dynamic.Core/Util/Cache/CacheContainer.cs
```diff
@@ -0,0 +1,14 @@
...
+
+internal struct CacheContainer<TValue> where TValue : notnull
+{
+    public CacheContainer(TValue value, DateTime expirationTime) 
+    {
+        Value = value;
+        ExpirationTime = expirationTime;
+    }
+
+    public TValue Value { get; }
```
StefH, line 11: Normally I put the properties and fields in the top of the file.

StefH (changes requested):
Thanks for the changes, only a few comments left. can you please take a look?

test/System.Linq.Dynamic.Core.Tests/Util/ThreadSafeSlidingCacheTests.cs
StefH: I understand. I was using an older version from xunit in the solution.
It's added here: https://github.com/xunit/xunit/issues/2609#issuecomment-1399639884

If you want; you upgrade all unit-test projects to use 2.5.0 which includes Assert.Fail

Else if this too much work, just keep the existing code.

test/System.Linq.Dynamic.Core.Tests/Util/Cache/ThreadSafeSlidingCacheTests.cs
```diff
@@ -0,0 +1,151 @@
...
+        var dateTimeUtilsMock = new Mock<IDateTimeUtils>();
+        dateTimeUtilsMock.SetupGet(d => d.UtcNow).Returns(UtcNow);
+
+        var cache = new ThreadSafeSlidingCache<int, string>(TimeSpan.FromMinutes(10),
+            dateTimeProvider: dateTimeUtilsMock.Object);
+
+        // Act
+        cache.AddOrUpdate(1, "one");
+
+        var r = dateTimeUtilsMock.Object.UtcNow.AddMinutes(11);
```
StefH, line 57: please use a better name instead of `r`

StefH (changes requested):

StefH:
@TWhidden 
Also, if you have solved my comment, you can "Resolve" this so that comment is closed in this issue page. Else I get lost reading all the open comments (which some are actually resolved) ;-)

Like:
image

src/System.Linq.Dynamic.Core/ParsingConfig.cs
```diff
@@ -234,4 +234,27 @@ public IQueryableAnalyzer QueryableAnalyzer
...
     /// </summary>
     public bool DisallowNewKeyword { get; set; } = false;
+
+    /// <summary>
+    /// Sets a Time-To-Live (TTL) for items in the constant expression cache to prevent uncontrolled growth. 
+    /// Items not accessed within this TTL will be expired, allowing garbage collection to reclaim the memory.
+    /// Default is 10 minutes.
+    /// </summary>
+    public TimeSpan ConstantExpressionSlidingCacheTimeToLive { get; set; } = TimeSpan.FromMinutes(10);
+
```
StefH, line 244: remove newline pls

src/System.Linq.Dynamic.Core/ParsingConfig.cs
```diff
@@ -234,4 +234,27 @@ public IQueryableAnalyzer QueryableAnalyzer
...
+
+    /// <summary>
+    /// Configures the minimum number of items required in the constant expression cache before triggering cleanup. 
+    /// This prevents frequent cleanups, especially in caches with few items. 
+    /// A default value of null implies that cleanup is always allowed to run, helping in timely removal of unused cache items.
+    /// </summary>
+    public int? ConstantExpressionSlidingCacheMinItemsTrigger { get; set; } = null;
+
```
StefH, line 252: remove newline pls

src/System.Linq.Dynamic.Core/ParsingConfig.cs
```diff
@@ -234,4 +234,27 @@ public IQueryableAnalyzer QueryableAnalyzer
     /// Default value is <c>false</c>.
     /// </summary>
     public bool DisallowNewKeyword { get; set; } = false;
+
+    /// <summary>
+    /// Sets a Time-To-Live (TTL) for items in the constant expression cache to prevent uncontrolled growth. 
+    /// Items not accessed within this TTL will be expired, allowing garbage collection to reclaim the memory.
+    /// Default is 10 minutes.
+    /// </summary>
+    public TimeSpan ConstantExpressionSlidingCacheTimeToLive { get; set; } = TimeSpan.FromMinutes(10);
```
StefH, line 243: In case we need a cache for other things, it's best to add a `CacheConfig` class with  3 values:
- TimeToLive
- MinItemsTrigger
- CleanupFrequency

And use this class here like:
``` c#
public CacheConfig? ConstantExpressionCacheConfig { get; set; }
```

TWhidden: I've included a default `CacheConfig` instance instead of opting for null support. This approach allows developers to directly modify the existing cache settings, eliminating the need to create a new configuration instance.

`public CacheConfig ConstantExpressionCacheConfig { get; set; } = new CacheConfig();`

TWhidden:
My bad on "Resolve conversation" - that would have helped me also when going through them all.

Was thinking about the new `CacheConfig` - atm, it's just a POCO - should it be passed as an alternate constructor item to the `ThreadSafeSlidingCache` object? Just cleans up its usage a little bit, since the config is only for it.

Also - `ThreadSafeSlidingCache` just sounds silly - should we just call it `SlidingCache` ?  Later, if there is another kind of cache, it wont have that same naming pattern?

StefH:
> Also - `ThreadSafeSlidingCache` just sounds silly - should we just call it `SlidingCache` ? Later, if there is another kind of cache, it wont have that same naming pattern?

SlidingCache is also fine for me.

StefH:
> Was thinking about the new `CacheConfig` - atm, it's just a POCO - should it be passed as an alternate constructor item to the `ThreadSafeSlidingCache` object? Just cleans up its usage a little bit, since the config is only for it.

If it makes it easier, yes you can do it.

StefH (approved):

StefH:
@TWhidden 
Thanks for the PR, I'll merge it and see if it correctly compiles on my system.

TWhidden:
Hey @StefH  - We need to look a little into this.  I think with high contention, we are creating some errors.  Here is our Sentry report stack trace.

I propose some minor changes:
1.) in `CleanupIfNeeded`, use the `Interlocked.Exchange`, use this with a finally block as an atomic operation to keep possible multiple entries into Cleanup()
2.) `_cache[key]` I think is throwing this error; It may have been removed under contention, move to a `TryGet`
3.) In the `TryGetValue` of `SlidingCache` - instead of checking the expire time here, if it does exist, we should just update / return the value, Its a micro-optimization, but we are lazy removing anyways.  if something is requesting it, maybe we should allow it to be returned, or make it a configuration to permit expired returns;
image

If you are ok with it, I would love to do another PR for it. Hopefully with less hassle this time to your process.

TWhidden:
Ill Open a new issue - and submit a PR;

Issue #769, opened 2024-02-02
Enhancements to SlidingCache<> for Improved Thread Safety and Performance


TWhidden:
In the latest update to the `SlidingCache<>` implementation, we addressed a critical oversight related to concurrent modifications and access patterns. Specifically, we identified that a cache key could be removed during enumeration, triggering a `KeyNotFoundException`. To address these concerns and enhance the cache's robustness and efficiency, the following adjustments were made:

1. **Atomic Locking with `Interlocked.CompareExchange`**: We introduced an atomic lock mechanism using `Interlocked.CompareExchange` to prevent simultaneous entries into the `Cleanup` method. This enhancement aims to significantly reduce contention and ensure thread-safe operations within the cache cleanup process.

2. **Deferred Cleanup Execution**: The `CleanupIfNeeded` method has been relocated to a finally block following the `TryGetValue` execution. This will ensure that if `PermitExpiredReturns` is enabled, there is no race to remove the item while accessing it.

3. **Support for Deterministic Element Caching**: A new parameter has been added to `SlidingCache<>`, enabling the caching of expired but deterministic elements. This feature allows expired objects to remain accessible before the Cleanup process has executed, preventing unnecessary removals and recreations. This capability is particularly beneficial for `ConstantExpressions` and is enabled by default to optimize performance for predictable data patterns.

4. **Configurable Expired Returns via `CacheConfig`**: To provide greater control over cache behavior, the `PermitExpiredReturns` setting has been incorporated into the `CacheConfig` class. This addition allows developers to manage the allowance of expired item returns based on their specific application requirements and caching strategies.

These enhancements are designed to bolster the `SlidingCache<>`'s performance, reliability, and flexibility, particularly in high-concurrency environments.

StefH:
@TWhidden 
Did you have time to verify this new PR?

TWhidden:
It's currently running in a beta instance without any known issues. I have not rolled it to prod yet.  I'll try to get that out this week or early next week. Sorry for the delay.

StefH:
Did you have time to test it?

TWhidden:
I have not pushed to production yet, however I did pull a memory snapshot of our demo/beta build which is an active system.  I am working on a release to prod right now.  FYI - this is using `9.3.9-preview-02` for reference.  So far, I have not had any error popup. 
[...]

PR #770, opened 2024-02-02
Enhancements to SlidingCache<> for Improved Thread Safety and Performance

TWhidden:
See #769 for description of changes;

### DIFF
--- a/src/System.Linq.Dynamic.Core/Util/Cache/CacheConfig.cs
+++ b/src/System.Linq.Dynamic.Core/Util/Cache/CacheConfig.cs
@@ -24,4 +24,13 @@ public class CacheConfig
     /// By default, cleanup occurs every 10 minutes.
     /// </summary>
     public TimeSpan CleanupFrequency { get; set; } = SlidingCacheConstants.DefaultCleanupFrequency;
+
+    /// <summary>
+    /// Enables returning expired cache items in scenarios where cleanup, running on a separate thread, 
+    /// has not yet removed them. This allows for the retrieval of an expired object without needing to 
+    /// clear and recreate it if a request is made concurrently with cleanup. Particularly useful 
+    /// when cached items are deterministic, ensuring consistent results even from expired entries.
+    /// Default true;
+    /// </summary>
+    public bool PermitExpiredReturns { get; set; } = true;
 }
--- a/src/System.Linq.Dynamic.Core/Util/Cache/CacheConfig.cs
+++ b/src/System.Linq.Dynamic.Core/Util/Cache/CacheConfig.cs
@@ -32,5 +32,5 @@ public class CacheConfig
     /// when cached items are deterministic, ensuring consistent results even from expired entries.
     /// Default true;
     /// </summary>
-    public bool PermitExpiredReturns { get; set; } = true;
+    public bool ReturnExpiredItems { get; set; } = true;
 }
--- a/test/System.Linq.Dynamic.Core.Tests/Util/Cache/SlidingCacheTests.cs
+++ b/test/System.Linq.Dynamic.Core.Tests/Util/Cache/SlidingCacheTests.cs
@@ -48,6 +48,7 @@ public class SlidingCacheTests
         var dateTimeUtilsMock = new Mock<IDateTimeUtils>();
         dateTimeUtilsMock.SetupGet(d => d.UtcNow).Returns(UtcNow);
 
+        // Arrange
         var cache = new SlidingCache<int, string>(TimeSpan.FromMinutes(10),
             dateTimeProvider: dateTimeUtilsMock.Object);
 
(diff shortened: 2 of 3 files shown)


src/System.Linq.Dynamic.Core/Util/Cache/CacheConfig.cs
```diff
@@ -24,4 +24,13 @@ public class CacheConfig
...
     public TimeSpan CleanupFrequency { get; set; } = SlidingCacheConstants.DefaultCleanupFrequency;
+
+    /// <summary>
+    /// Enables returning expired cache items in scenarios where cleanup, running on a separate thread, 
+    /// has not yet removed them. This allows for the retrieval of an expired object without needing to 
+    /// clear and recreate it if a request is made concurrently with cleanup. Particularly useful 
+    /// when cached items are deterministic, ensuring consistent results even from expired entries.
+    /// Default true;
+    /// </summary>
+    public bool PermitExpiredReturns { get; set; } = true;
```
StefH, line 35: Maybe:
`ReturnExpiredItems` ?

test/System.Linq.Dynamic.Core.Tests/Util/Cache/SlidingCacheTests.cs
```diff
@@ -63,6 +63,27 @@ public void SlidingCache_TestExpire()
         }
     }
 
+    [Fact]
+    public void SlidingCache_TestExpiredReturn()
+    {
+        var dateTimeUtilsMock = new Mock<IDateTimeUtils>();
```
StefH, line 69: Add `// Arrange` as comment

StefH, line 69: Use "SetupSequence" if possible instead of `SetupGet`.

Example:
 
## Sequences

What about the times you expect a method to be called three times, each with different results? You're looking for Moq's SetupSequence method, which can be used in-place of the Setup method. As seen in the following code snippet, it's possible to stack your mock with several expectations that you're expecting to happen in a set sequence order.

``` c# 
// Arrange
var mock = new Mock<IDateTimeProvider>();
mock.SetupSequence(m => m.DayOfWeek())
    .Returns(DayOfWeek.Sunday)
    .Returns(DayOfWeek.Tuesday)
    .Returns(DayOfWeek.Saturday);

// Act
var result = 
    ExampleClass.ChecksSequenceIsCorrect(
        mockedObject: mock.Object);

// Assert
Assert.That(result, Is.EqualTo(true));
mock.VerifyAll();
```

See https://www.codemag.com/Article/2305041/Using-Moq-A-Simple-Guide-to-Mocking-for-.NET

test/System.Linq.Dynamic.Core.Tests/Util/Cache/SlidingCacheTests.cs
```diff
@@ -63,6 +63,27 @@ public void SlidingCache_TestExpire()
...
+        var cache = new SlidingCache<int, string>(TimeSpan.FromMinutes(10),
+            dateTimeProvider: dateTimeUtilsMock.Object, permitExpiredReturns: true);
+
+        // Act
+        cache.AddOrUpdate(1, "one");
+
+        var newDateTime = dateTimeUtilsMock.Object.UtcNow.AddMinutes(11);
+        dateTimeUtilsMock.SetupGet(d => d.UtcNow).Returns(newDateTime);
+
+        if (!cache.TryGetValue(1, out var value))
```
StefH, line 81: Use FluentAssertions to make a a one liner:
```
cache.TryGetValue(1, out var value).Should().BeFalse($"Expected to not find the value, but found {value}");
```

StefH (changes requested):

test/System.Linq.Dynamic.Core.Tests/Util/Cache/SlidingCacheTests.cs
TWhidden: I'm facing challenges with integrating the Mock framework for unit testing, particularly with the `SlidingCache` component:
- I'm having difficulty setting it up without encountering exceptions. Despite several attempts, I can't seem to find a configuration that works.
- My current understanding is that when using `.Returns` with Mock, it's supposed to return the next item in a sequence each time a `DateTime` call is made. However, I'm concerned about the timing of cleanup operations. Specifically, if a cleanup operation runs on a ThreadPool thread, it might execute either before or after other operations, leading to non-deterministic behavior. I'm not sure if I'm missing something in my implementation approach. As a workaround, I've been manually adjusting the `DateTime` to force expiration of items in the cache, which bypasses the need to manage the internal timing.

Could you help me out with one of the test cases?

StefH: In this case, just add a small comment to the unit test why it is like this. No need to change the Moq code.

TWhidden: I see on the other tests you refactored last time, you had added  `// move the time forward.` so I just used that for consistency.  Included in that last check-in.  I'll mark this as Resolved.

StefH (commented):
just add a comment to the unit test file

StefH (approved):
.

StefH:
I did a small fix on the unit tests.
https://github.com/zzzprojects/System.Linq.Dynamic.Core/commit/e6bc36e7d466f4664318f957bba77729ff19954c

And I'll release a new preview soon.

PR #797, opened 2024-04-15
Add Json support: check return types of method

paule96:
Hi @StefH,

I fixed the single and first methods that accordingly to the [C# docs](https://learn.microsoft.com/en-us/dotnet/api/system.linq.enumerable.first?view=net-8.0) should return an invalid operation with `Sequence contains no elements`.


StefH:
Thanks

PR #824, opened 2024-06-25
Fixed: Incorrect Handling of Qualifiers in ConstantExpressionHelper

RenanCarlosPereira:
**Description:**
There is an issue in the `ExpressionPromoter` class where the qualifier in the numeric literal expressions is not correctly handled, leading to incorrect type promotions. This is due to the way qualifiers are stored and retrieved from the `ConstantExpressionHelper`.

**Steps to Reproduce:**
1. Create an instance of `ParsingConfig`.
2. Use `NumberParser` to parse a numeric literal with a qualifier.
3. Attempt to promote the parsed expression to a different type using `ExpressionPromoter`.

**Example 1:**

```csharp
[Fact]
public void promoter()
{
    // Assign
    var parsingConfig = new ParsingConfig();
    var numberParser = new NumberParser(parsingConfig);
    var promoter = new ExpressionPromoter(parsingConfig);

    var doubleExpression = numberParser.ParseRealLiteral("10.5d", 'd', true);

    // Act & Assert
    var promotedExpression = promoter.Promote(doubleExpression, typeof(decimal), true, true);

    Assert.NotNull(promotedExpression);
    Assert.Equal(typeof(decimal), promotedExpression.Type);
}
```

**Example 2:**

```csharp
[Fact]
public void promoteDoubleToDecimal()
{
    // Assign
    var parsingConfig = new ParsingConfig();
    var numberParser = new NumberParser(parsingConfig);
    var promoter = new ExpressionPromoter(parsingConfig);

    var doubleExpression = numberParser.ParseRealLiteral("123.456d", 'd', true);

    // Act & Assert
    var promotedExpression = promoter.Promote(doubleExpression, typeof(decimal), true, true);

    Assert.NotNull(promotedExpression);
    Assert.Equal(typeof(decimal), promotedExpression.Type);
}
```

**Expected Behavior:**
The `Promote` method should correctly handle the qualifier and convert the numeric literal to the specified target type.

**Actual Behavior:**
The promotion fails, and the conversion does not occur as expected due to the incorrect handling of the qualifier in the `ConstantExpressionHelper`.

**Additional Information:**
This issue affects applications that rely on dynamic LINQ queries where numeric-type promotions with qualifiers are necessary.

**Proposed Solution:**
Ensure that the `CreateLiteral` in `ConstantExpressionHelper` correctly handles qualifiers without the prefix when creating and retrieving constant expressions.


RenanCarlosPereira:
@StefH  could you review it, then if possible we could update the rules engine 😅

RenanCarlosPereira:
the bug related is in the rules engine, I actually need to open a bug over there and reference it here: 
because the bug itself its not in the rules engine, but this PR will solve the issue and we will have to upgrade the package: 
https://github.com/microsoft/RulesEngine/issues/604#issuecomment-2185151056


I am unsure why GitHub actions are not triggering, merge is done!

PR #833, opened 2024-07-29
Update DefaultDynamicLinqCustomTypeProvider constructor to be compatible with microsoft/RulesEngine

StefH:


aligneddev (approved):
That looks like it will handle the issue to me.

StefH:
> That looks like it will handle the issue to me.

To be 100% sure, I did create a preview (System.Linq.Dynamic.Core.1.4.4-preview-01) which can be downloaded from MyGet:
https://www.myget.org/F/system-linq-dynamic-core/api/v3/index.json

Can you test this?

nmadland:
Yes will do. Thank you.

nmadland:
We should be back in business. Thank you.

StefH:
Fixing SonarCloud is not needed, because I actually introduced a obsolete and not-used method ;-)

nmadland:
Yes it does. Thank you.

nmadland:
Any idea when a new version will be published to nuget?

StefH:
> Any idea when a new version will be published to nuget?

probably in a few days

PR #870, opened 2025-01-22
Fix CVE-2024-51417

StefH:
- remove 'object' from PredefinedTypes
- refactor DefaultDynamicLinqCustomTypeProvider so that only classes with DynamicLinqType annotation are resolved


test/System.Linq.Dynamic.Core.Tests/SecurityTests.cs
```diff
@@ -64,4 +62,129 @@ public void DynamicExpressionParser_ParseLambda_IllegalMethodCall_ThrowsExceptio
...
+        }.AsQueryable();
+
+        // Act
+        Action action = () => queryable.Select(config, selector);
+
+        // Assert
+        action.Should().NotThrow();
+    }
+
+    [Theory(Skip = "Bug: Accessing static members on instance class is not supported")]
```
Tasteful, line 148: Why are this test skipped? Should the test case not be reverted to ensure that it throws?

StefH: I investigated this, and I think was is an existing bug. But I'll take a look if I can fix that also.

StefH: These tests also fail in the current master:
``` c#
[Theory]
[InlineData("Prop", "TestProp")]
[InlineData("Field", "TestField")]
public void Parse_StaticPropertyOrField_In_NonStaticClass1(string name, string value)
{
    // Arrange
    var ints = new int[1].AsQueryable();

    // Act
    var result = ints.Select<string>($"new {typeof(NonStaticClassExample)}().{name}").First();

    // Assert
    Assert.Equal(value, result);
}

[Theory]
[InlineData("Prop", "TestProp")]
[InlineData("Field", "TestField")]
public void Parse_StaticPropertyOrField_In_NonStaticClass2(string name, string value)
{
    // Arrange
    var ints = new NonStaticClassExample[] { new NonStaticClassExample() }.AsQueryable();

    // Act
    var result = ints.Select<string>(name).First();

    // Assert
    Assert.Equal(value, result);
}

[DynamicLinqType]
public static class StaticClassExample
{
    public static string Prop { get; set; } = "TestProp";

    public static string Field = "TestField";
}

[DynamicLinqType]
public class NonStaticClassExample
{
    public static string Prop { get; set; } = "TestProp";

    public static string Field = "TestField";
}
```

So I will create a new bug for this and solve it with a new PR, in order to keep this PR only focused on the CVE.

mhubis:
ResolveTypeBySimpleName's call to FindTypesMarkedWithDynamicLinqTypeAttribute is missing the union with AdditionalTypes. ResolveType makes the same call to FindTypesMarkedWithDynamicLinqTypeAttribute but does union with AdditionalTypes. 

Can this be fixed?

StefH:
> ResolveTypeBySimpleName's call to FindTypesMarkedWithDynamicLinqTypeAttribute is missing the union with AdditionalTypes. ResolveType makes the same call to FindTypesMarkedWithDynamicLinqTypeAttribute but does union with AdditionalTypes.
> 
> Can this be fixed?

Good point. I'm fixing it with: https://github.com/zzzprojects/System.Linq.Dynamic.Core/pull/896

PR #871, opened 2025-01-23
Use MemberInfo to detect methods that should be allowed. 

Tasteful:
Add Equals and ToString from object to allowed as default.

Note: Was not able to run the code locally because of missing workloads in Visual Studio.

@StefH Can this be a way forward to make the object.Equal and object.ToString to work as well?


StefH:
@Tasteful 

At this moment, also calling `ToString()` or `Equals` on an _"safe"_ primitive type like `int` or `bool` is not allowed when `AllowEqualsAndToStringMethodsOnObject` is **not** set to `true`.

I think it's safe to update the logic for the PredefinedMethodsHelper so that the `ToString()` or `Equals` methods can be called on these primitive type. What do you think?

StefH:
And what about the `ToString` which takes a IFormatProvider and a string + IFormatProvider ?

Tasteful:
I think overloads as well should be accessible, I cant directly think about a way to use the to inject my own code or extract information that not are on the object that i'm working with.

PR #875, opened 2025-01-23
Add option to ParsingConfig to allow the Equals and ToString methods on object

StefH:


StefH:
@Tasteful

I used your idea for allowing the Equals and ToString methods with a config setting.

Note that for security reasons, it's not allowed by default.

StefH:
Hello @ChrisGilbertInfor,

Sorry, I cannot comment on the security implications of enabling this setting for your applications.

About **int.ToString()**, I think it's possible and also safe to allow `ToString `and `Equals `on these types:


typeof(bool)
typeof(char)
typeof(string)
typeof(sbyte)
typeof(byte)
typeof(short)
typeof(ushort)
typeof(int)
typeof(uint)
typeof(long)
typeof(ulong)
typeof(float)
typeof(double)
typeof(decimal)
typeof(DateTime)
typeof(DateTimeOffset)
typeof(TimeSpan)
typeof(Guid)
typeof(Uri)
typeof(DateOnly)
typeof(TimeOnly)

I'll think about this.

PR #905, opened 2025-03-23
Fix: Add Fallback in ExpressionPromoter to Handle Cache Cleanup in ConstantExpressionHelper

RenanCarlosPereira:
### 🧩 Problem Summary

In multi-threaded scenarios, a race condition can occur due to the internal behavior of `SlidingCache` used in `ConstantExpressionHelper`. Specifically:

- When a constant is added via `CreateLiteral(value, text)`, both `_expressions` and `_literals` caches are populated.
- Later, if `_literals` is cleaned up (due to TTL or cleanup frequency), `ExpressionPromoter.Promote()` will fail to retrieve the text representation of the constant.
- As a result, **numeric promotion fails**, causing runtime exceptions like:

```
Operator '<' incompatible with operand types 'Decimal' and 'Double'
```

This issue was especially noticeable when comparing expressions like:
```csharp
(x.A / x.B) < 0.40
```

### ✅ What This PR Does

This PR adds a **safe fallback mechanism** in `ExpressionPromoter.Promote()`:
- If `_constantExpressionHelper.TryGetText(...)` fails (because `_literals` was cleaned),
- We now fallback to `ce.Value?.ToString()` to obtain a text representation for parsing.
- Type promotion then continues as expected using the parsed value.

This ensures the parser remains **robust and consistent**, even when cache entries are removed in the background.

### 🧪 How We Reproduced the Issue

A unit test was created to simulate this scenario by:
1. Adding a constant via `CreateLiteral(...)`.
2. Manually removing the corresponding `_literals` entry.
3. Calling `Promote()` after cleanup.
4. Verifying that promotion fails without a fallback — and succeeds after this fix.

### 📈 Impact

- Fixes an intermittent failure that can happen in real-world concurrent scenarios.
- Makes `ExpressionPromoter` safer and more fault-tolerant.
- Enables continued parsing even if internal caches are cleaned up.

---

### 🚀 Ready for Review

Let me know if there are preferences for making the fallback smarter (e.g. re-adding the literal to cache or applying type suffixes). This version keeps it simple and safe.


asulwer (approved):

StefH (changes requested):
1. When I just copy this test to the master branch, it does not fail, can you double check if this test actually tests this exact scenario?

2. The fix is limited to 1 place, but the ConstantExpressionHelper is used in multiple places.
I was just wondering, would setting ReturnExpiredItems to true also solve this issue in all places?

RenanCarlosPereira:
1.  It's hard to simulate the problem because of race conditions. 
Maybe if you increase the `CancellationTokenSource ` delay, you will see it failing.  I tried changing to master and adding the test, and for me, the tests are not passing. 

2. I double-checked, and it is testing this scenario.
    Setting up the ReturnExpiredItems to true won't solve the problem because the method that is cleaning the dictionary is the `Cleanup `method, which doesn't take into consideration the ReturnExpiredItems.
    
   
3. If you want to get together on a quick call, I could explain the situation 😁

RenanCarlosPereira:
any updates on this? thanks

grichards23:
Any idea when this fix will be merged?

RenanCarlosPereira:
Im waiting for the maintainers, while we wait I set up the cache to 10 years, thats my workaround. 🤣

Sent from Outlook for iOS<https://aka.ms/o0ukef>
________________________________

StefH:
@asulwer 
I did confirm that the unit tests do fail when running on old code, and are running correct on this new code.

I'll try to see if I can verify once more, and then merge.

RenanCarlosPereira:
Any updates on this review?

RenanCarlosPereira:
any updates on this issue?

test/System.Linq.Dynamic.Core.Tests/Parser/ExpressionPromoterTests.cs
```diff
@@ -1,56 +1,132 @@
...
-    }
+        public class SampleDto
+        {
+            public Guid data { get; set; }
+        }
 
-    private readonly Mock<IExpressionPromoter> _expressionPromoterMock;
-    private readonly Mock<IDynamicLinqCustomTypeProvider> _dynamicLinkCustomTypeProviderMock;
+        private readonly Mock<IExpressionPromoter> _expressionPromoterMock;
+        private readonly Mock<IDynamicLinkCustomTypeProvider> _dynamicLinkCustomTypeProviderMock;
```
StefH, line 22: Can you use `IDynamicLinqCustomTypeProvider` and also rename `_dynamicLinkCustomTypeProviderMock` to `_dynamicLinqCustomTypeProviderMock` ?

StefH (changes requested):

src/System.Linq.Dynamic.Core/Parser/ExpressionPromoter.cs
```diff
@@ -50,7 +50,12 @@ public ExpressionPromoter(ParsingConfig config)
             }
             else
             {
-                if (_constantExpressionHelper.TryGetText(ce, out var text))
+                if (!_constantExpressionHelper.TryGetText(ce, out var text))
```
StefH, line 53: Instead of adding this logic here, I would be better I think to add this to the helper:
``` c#
public bool TryGetText(ConstantExpression expression, [NotNullWhen(true)] out string? text)
{
    if (_literals.TryGetValue(expression, out text))
    {
        return true;
    }

    if (expression.Value is string stringValue)
    {
        text = stringValue;
        return true;
    }

    return false;
}
```

StefH (changes requested):
please take a look at my comments

RenanCarlosPereira:
@StefH I updated to use the IDynamicLinqCustomTypeProvider, the abstraction I think we could leave to another PR, I can do it later on if you want.

PR #940, opened 2025-07-29
Lazily load dynamic linq types

### LINKED ISSUE #939: CPU Spike at startup
When calling for the first time `OrderBy`, it seems like [it iterates on all types of all loaded assemblies](https://github.com/zzzprojects/System.Linq.Dynamic.Core/blob/53361d6cac16affbc70666966d8a373992fc1282/src/System.Linq.Dynamic.Core/CustomTypeProviders/DefaultDynamicLinqCustomTypeProvider.cs#L104-L108). That does not scale and consumes 20% of the CPU for ~1 minute during the startup of my application.

Additionally, it seems like there is no synchronization around this operation so all threads are racing.

Image

Is there an opportunity to lazily load the type metadata individually?

verdie-g:
Partially fixes #939.

I found two code paths that requires iterating on all the types of the world:
1. [KeywordsHelper.TryGetValue](https://github.com/zzzprojects/System.Linq.Dynamic.Core/blob/53361d6cac16affbc70666966d8a373992fc1282/src/System.Linq.Dynamic.Core/Parser/KeywordsHelper.cs#L127-L132)
2. [MethodFinder.FindMethod](https://github.com/zzzprojects/System.Linq.Dynamic.Core/blob/53361d6cac16affbc70666966d8a373992fc1282/src/System.Linq.Dynamic.Core/Parser/SupportedMethods/MethodFinder.cs#L139)

In both cases, the "custom types" are not used first so there is a chance that they are never used and don't need to be loaded.

### DIFF
--- a/src/System.Linq.Dynamic.Core/Compatibility/Lazy.cs
+++ b/src/System.Linq.Dynamic.Core/Compatibility/Lazy.cs
@@ -1,0 +1,40 @@
+using System.Threading;
+
+#if NET35
+namespace System
+{
+    internal class Lazy<T>
+    {
+        private readonly Func<T> _valueFactory;
+        private readonly object _lock;
+
+        private T? _value;
+        private bool _valueCreated;
+
+        public Lazy(Func<T> valueFactory, LazyThreadSafetyMode mode)
+        {
+            _valueFactory = valueFactory;
+            _lock = new object();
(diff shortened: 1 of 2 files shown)


verdie-g:
@StefH what do you think of that PR?

PR #944, opened 2025-09-09
Add AllowAccessToNonPublicMembers property to ParsingConfig

pirumar:
This new property allows the parser to consider non-public members during property/field lookups when access is permitted. Updated ExpressionParser to utilize this configuration for member visibility during lookups.


StefH:
@pirumar 

Can you also add unit tests?

PR #953, opened 2025-10-25
Fixed adding Enum and integer

StefH:


test/System.Linq.Dynamic.Core.Tests/QueryableTests.Select.cs
```diff
@@ -202,6 +202,32 @@ public void Select_Dynamic_Add_Strings()
...
+        public void Select_Dynamic_Add_DayOfWeekEnum_And_Integer()
+        {
+            // Arrange
+            var range = new DayOfWeek[] { DayOfWeek.Monday };
+
+            // Act
+            var rangeResult = range.AsQueryable().Select("it + 1");
+
+            // Assert
+            Assert.Equal(range.Select(x => x + 1).ToArray(), rangeResult.Cast<DayOfWeek>().ToArray());
```
glopesdev: @StefH I believe this is the misleading unit test: `rangeResult.Cast<DayOfWeek>()` will work, since indeed it is possible to convert the resulting integer to the enum. However, the type inferred from the expression is `int` rather than `DayOfWeek`.

Issue #967, opened 2025-12-03
LINQ methods like Min, Max and Sum remain case sensitive when working with DateTime


victorr99:
We observed the strange behavior that the LINQ methods like `Min`,  `Max` and `Sum` remain case sensitive when working with `DateTime` properties. When working with integers for example, thay are indeed case insensitive. We have the option `IsCaseSensitive` set to false.

Code to reproduce:
```cs
 var models = new Model[]
 {
     new Model { Date = new DateTime(2025, 12, 10, 12, 0, 0) },
     new Model { Date = new DateTime(2025, 8, 10, 12, 0, 0) }
 };

 var parameter = Expression.Parameter(typeof(Model[]), "models");

 var parser = new ExpressionParser(
     [parameter],
     "models.MIN(x => x.Date)", // or min, MAX, max
     [],
     new ParsingConfig
     {
          IsCaseSensitive = false
     });

 var expression = parser.Parse(typeof(DateTime));
 var labmda = Expression.Lambda<Func<Model[], DateTime>>(expression, parameter);
 var method = labmda.Compile();

 var result = method(models);
 Console.Write(result);

public class Model
{
    public DateTime Date { get; set; }
}
```

Exception:
```
Unhandled exception. System.InvalidOperationException: No generic method 'MIN' on type 'System.Linq.Enumerable' is compatible with the supplied type arguments and arguments. No type arguments should be provided if the method is non-generic.
   at System.Linq.Expressions.Expression.FindMethod(Type type, String methodName, Type[] typeArgs, Expression[] args, BindingFlags flags)
   at System.Linq.Expressions.Expression.Call(Type type, String methodName, Type[] typeArguments, Expression[] arguments)
   at System.Linq.Dynamic.Core.Parser.ExpressionParser.TryParseEnumerable(Expression instance, Type enumerableType, String methodName, Type type, Expression[]& args, Expression& expression)
   at System.Linq.Dynamic.Core.Parser.ExpressionParser.ParseMemberAccess(Type type, Expression expression, String id)
   at System.Linq.Dynamic.Core.Parser.ExpressionParser.ParsePrimary()
   at System.Linq.Dynamic.Core.Parser.ExpressionParser.ParseUnary()
   at System.Linq.Dynamic.Core.Parser.ExpressionParser.ParseArithmetic()
   at System.Linq.Dynamic.Core.Parser.ExpressionParser.ParseAdditive()
[...]

PR #968, opened 2025-12-05
LINQ methods respect the configured casing

victorr99:
The ExpressionParser was updated with a new method allowing case insensitive compares whenever this was configured in the ParserConfig.

The method `CanonicalContains` performs a case insensitive check whenever `IsCaseSensitive` was set to false. It also updates the provided `needle` parameter to the actual found value. Thus a match on `max` will be translated to `Max`.

This fixes issue #967

### DIFF
--- a/src/System.Linq.Dynamic.Core/Parser/ExpressionParser.cs
+++ b/src/System.Linq.Dynamic.Core/Parser/ExpressionParser.cs
@@ -2139,7 +2139,7 @@ public class ExpressionParser
         // Create a new innerIt based on the elementType.
         var innerIt = ParameterExpressionHelper.CreateParameterExpression(elementType, string.Empty, _parsingConfig.RenameEmptyParameterExpressionNames);
 
-        if (CanonicalContains(["Contains", "ContainsKey", "Skip", "Take"], ref methodName))
+        if (Contains(["Contains", "ContainsKey", "Skip", "Take"], methodName))
         {
             // For any method that acts on the parent element type, we need to specify the outerIt as scope.
             _it = outerIt;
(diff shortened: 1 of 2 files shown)


victorr99:
@StefH can you have a look again at the PR?

victorr99:
@StefH any updates?

victorr99:
The only compatability issue I can think of is that a custom enumerable object is being used which specifies `Min` and `Max` methods on its own. Think of the following class:

```cs
public class CustomEnumerable<T> : IEnumerable<T>
{
    public int Max { get; set; }

    public int Min()
    {
    }
}
```

The `Max` property can easily be distinguished from the method because the parantheses makes it a method, unless it's a delegate like `Func<int>`. For the `Min` method and (the `Func<int> Max` property), the implementation of the CustomEnumerable should be choosen.
I have never needed to use a custom enumerable class, but I can imagine that there are use cases for that. I can add extra tests which checks that the methods of the custom enumerable are choosen over the extension methods, however I don't think proper support for custom enumerables has ever existed. The following code returns `15`, but it should return `0`:

```cs
public class CustomEnumerable<T>(IEnumerable<T> inner) : IEnumerable<T>
{
    public T Max { get; set; }

    public DateTime Min()
    {
        return default;
    }

    public IEnumerator<T> GetEnumerator()
    {
        return inner.GetEnumerator();
    }

    IEnumerator IEnumerable.GetEnumerator()
    {
        return GetEnumerator();
    }
}

var parameter = Expression.Parameter(typeof(CustomEnumerable<int>), "list");

var parser = new ExpressionParser(
    [parameter],
    "list.Min()",
    [],
    new ParsingConfig { IsCaseSensitive = false });

var expression = parser.Parse(typeof(int));
var lambda = Expression.Lambda<Func<CustomEnumerable<int>, int>>(expression, parameter);
var method = lambda.CompileFast();

Console.WriteLine(method(new CustomEnumerable<int>([15])));
```

So, I assume fixing it the right way introduces another backward compatability issue, or the CustomEnumerable example is a new bug.

By the way, you asked for extra tests which covers the `Min` and `Max` properties on a class. I don't think it adds extra value to the other test cases since LING is being applied to an enumerable of objects. It is not an extension for that object/class.

Issue #983, opened 2026-06-25
How to IEnumerable<string>.Contains(x, StringComparer.OrdinalIgnoreCase) ?


ahdung:
<img width="739" height="567" alt="Image" src="https://github.com/user-attachments/assets/3b38052f-7d66-4e50-8ef2-170dfd76fe47" />

ahdung:
I'm not use for db query, just linq to object in memory.
I tried `Contains(name, "OrdinalIgnoreCase")` and `Contains(name, StringComparison.OrdinalIgnoreCase)`, no exception, but result not correct, still case senstive.

<img width="720" height="413" alt="Image" src="https://github.com/user-attachments/assets/b7b2f37d-4419-417c-97bd-2defff0feb4a" />

<img width="717" height="455" alt="Image" src="https://github.com/user-attachments/assets/74542ccd-dd2a-452a-8922-732172371282" />

PR #984, opened 2026-06-27
By default, also support WellKnownTypes like StringComparer

StefH:


StefH:
https://github.com/zzzprojects/System.Linq.Dynamic.Core/issues/983

